PR #2, opened 2016-08-08
Fix a typo in DESCRIPTION

haozhu233:


hadley:
That's not a typo, unless you think my middle name is `hadley@rstudio.com` ;)

PR #5, opened 2016-08-09
enable packrat

ellisvalentiner:


hadley:
I prefer not to use packrat on development packages.

PR #9, opened 2016-08-09
Add fct_shift() source and README.

rudazhan:
`fct_shift()` shifts the order of levels to the left/right. 
This is inspired by [r4ds modeling excercise](http://r4ds.had.co.nz/model-building.html#exercises-64) number 8.


hadley:
I'd build this on top of something lower-level like:

```R
shift <- function(x, n) {
  idx <- ((seq_along(x) - 1) + n) %% length(x) + 1
  x[idx]
}
shift(1:10, -1)
shift(1:10, 0)
shift(1:10, 1)
```

R/shift.R
```diff
@@ -0,0 +1,20 @@
...
+#' @examples
+#' x <- factor(weekdays(Sys.Date()), levels = c("Sunday", "Monday", "Tuesday",
+#'             "Wednesday", "Thursday", "Friday", "Saturday"), ordered = TRUE)
+#' fct_shift(x)
+#' fct_shift(x, left_by = 1)    # The same
+fct_shift <- function(f, left_by = 1L) {
+  stopifnot(length(left_by) == 1L)
+  mark <- left_by %% nlevels(f)
+  new_levels <- levels(f)[c(seq(mark + 1, nlevels(f)), seq_len(mark))]
+  factor(f, levels = new_levels, ordered = TRUE)
```
hadley: I don't think you should set `ordered` here - the default is to inherit from `f`

rudazhan:
I agree that level ordering shouldn't be forced. 

The shift function would be easier to read if it looks like:
```R
shift <- function(x, n) {
  idx <- (seq_along(x) + n) %% length(x)
  idx[idx == 0] <- length(x)
  x[idx]
}
```
But I used your version since it is not exported.

hadley:
Thanks!

PR #11, opened 2016-08-10
Make reorder to work with unused factor levels

### LINKED ISSUE #10: fct_reorder fails with unused factor levels
If a factor to reorder contains unused levels, the ` fct_reorder` fails to reorder correctly and will return NA values.

```
dat <- data.frame(f = factor(c("a", "b", "c", "d"), 
                            levels = c("c", "b", "a", "d")), 
                 x = c(1, 2, 3, 4)) %>% 
  filter(f != "b") %>%
  mutate(f = fct_reorder(f, x)) 

dat$f

## [1] a    c    <NA>
## Levels: b c a

```

The reason is a `factor(f)` inside tapply, which drop unused levels, but `factor()` that returns uses levels from original f.

jhuovari:
fixes #10 

This was fixed for `reorder2` in 854c7ea6e6452ffd291e9db1056f1b2574593738 but not for `reorder`.

I don't know whether  `as.factor()` is necessary at all, as `check_factor(f)` will ensure that f is a factor, but this is in line with `reorder2`.

### DIFF
--- a/R/reorder.R
+++ b/R/reorder.R
@@ -36,7 +36,7 @@ fct_reorder <- function(f, x, fun = median, ...) {
   f <- check_factor(f)
   stopifnot(length(f) == length(x))
 
-  summary <- tapply(x, as.factor(f), fun, ...)
+  summary <- tapply(x, f, fun, ...)
   if (!is.numeric(summary)) {
     stop("`fun` must return a single number per group", call. = FALSE)
   }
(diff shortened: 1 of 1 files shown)


hadley:
Since `check_factor()` is there now, it would be better to remove `as.factor`/`factor` from both functions. Would you mind doing that?

jhuovari:
Now they are removed from both functions.

hadley:
Thanks!

PR #36, opened 2016-09-07
Adds package level doc

jennybc:
... and 2 other minor housekeeping fixes that I think are OK to bundle: added `doc/` to Rbuildignore to clear a NOTE and fixed a typo. Confess I didn't do whatever actually populates `docs/` and I imagine the package-level help should appear there as well.


R/forcats-package.r
```diff
@@ -0,0 +1,10 @@
+#' forcats.
+#'
+#' Helpers for reordering factor levels (including moving specified levels to
+#' front, ordering by first appearance, reversing, and randomly shuffling), and
+#' tools for modifying factor levels (including collapsing rare levels into
+#' other, anonymising, and manually recoding).
+#'
+#' @name forcats
+#' @docType package
+NULL
```
hadley: Can you please use `"_PACKAGE"` instead of NULL?  Then you don't need `@docType` and some other stuff should get filled in automatically.

jennybc: I think this presumes the dev version of roxygen2; I was using CRAN. You still want?

hadley:
Thanks!

PR #47, opened 2016-10-18
fct_paste

MarcusWalz:
Function that allows "pasting" two factors together. Sometimes useful for plotting.

```
> fct_paste(c("a", "b", "c"),fct_rev(c("x","y","z")))
[1] "a z" "b z" "c z" "a y" "b y" "c y" "a x" "b x" "c x"
[1] a x b y c z
Levels: a z b z c z a y b y c y a x
```


R/paste.R
```diff
@@ -0,0 +1,24 @@
+#' "Paste" together factors.
+#'
+#' Order is cartesian product of all levels.
+#'
+#' This is sometimes useful when plotting a factor
```
krlmlr: The third sentence goes into "Details", is this intended? How is your implementation different from base::interaction()?

MarcusWalz: Good call. I would recommend implementing it using `base::interaction()` if this feature is worthwhile.

hadley:
I don't understand why you want this, and I don't think `paste` is the right name, since it acts more like grid.  Can you please file an issue with a motivating example?

MarcusWalz:
I probably use this once a month and it generally happens in meta analysis where I need to paste together columns while preserving chronological ordering. And `base::interaction()` does exactly what I want. But in case the issue appears again, here is my use case:

Suppose `model` which is a factor with names prognostic models for a disease in chronological order of publication. And `endpoint` is a factor describing endpoints of the model (a model can have more than one) in the order `c("Recurrence", "Progression", "Death")`. I want to combine `endpoint` and `model` together so Recurrence shows up first and Death shows up last and models are listed in original ordering. In dplyr I don't need to combine at all, I can use ``arrange(endpoint, model)`` or ``group_by(endpoint, model)``. In ggplot2 I'm generally in a facet scenario and ``facet_wrap(~ endpoint + model, ncol=1)`` has the desired behavior. However, if I need to step outside the comfort zone that is the tidyverse, I'll sometimes need to untidy the data and merge the two factors together using ``base::interaction(endpoint, model, sep=" ")``.

PR #49, opened 2016-10-31
Quick fix for link on forcats README pages

colinbrislawn:
(I could also be wrong. Maybe `forcats` needs `readr`.)


hadley:
Thanks!

PR #50, opened 2016-11-01
New fct_relabel()

krlmlr:
to change factor labels programmatically.

@hadley: PTAL.


R/relabel.R
```diff
@@ -0,0 +1,41 @@
...
+#'   convert_income)
+#' fct_count(rincome2)
+fct_relabel <- function(f, fun, ...) {
+  if (!is.function(fun)) {
+    stop("expected function, got ", class(fun), call. = FALSE)
+  }
+
+  old_levels <- levels(f)
+  new_levels <- fun(old_levels, ...)
+  if (!is.character(new_levels)) {
```
hadley: I think it would be better to use `lvls_revalue()` here - the error messages won't be quite as good, but it keeps all the logic in one place.

R/relabel.R
```diff
@@ -0,0 +1,41 @@
...
+#' }
+#'
+#' convert_income(levels(gss_cat$rincome))
+#'
+#' rincome2 <- fct_relabel(gss_cat$rincome,
+#'   convert_income)
+#' fct_count(rincome2)
+fct_relabel <- function(f, fun, ...) {
+  if (!is.function(fun)) {
+    stop("expected function, got ", class(fun), call. = FALSE)
```
hadley: Can you please start error messages with a capital letter? 

Also need to think about what happens when class has multiple values.

R/relabel.R
```diff
@@ -0,0 +1,41 @@
...
+#'   is_range <- grepl(regex, x)
+#'   num_income <- as.numeric(gsub(regex, "\\1", x[is_range]))
+#'   num_income <- trunc(num_income / 5000) * 5000
+#'   x[is_range] <- paste0("Gt $", num_income)
+#'   x
+#' }
+#'
+#' convert_income(levels(gss_cat$rincome))
+#'
+#' rincome2 <- fct_relabel(gss_cat$rincome,
```
hadley: Should be one line

krlmlr:
Addressed all remarks and rebased, lvls_relabel() now shows a richer error message.

hadley:
Thanks!

Issue #84, opened 2017-02-26
fct_recode(), rename factor level to f doesn't work


Tazinho:
The following gives an error, since f is also the name of the data argument.
Is there a recommended workaround for these situations?

> a <- factor(letters[1:5])
> forcats::fct_recode(a, f= "a", h = "b")
Error: Each input to fct_recode must be a single named string. Problems at positions: 1

QuentinRoy:
Same issue. I believe the problem is that 'f' is recognized as false here:
```r
  is_ok <- function(x) is.character(x) && length(x) == 1
  ok <- vapply(levels, is_ok, logical(1))
```

hadley:
Thanks for the bug report! I'll attack this systematically as part of #110

gabrielburcea:
i got this problem for fct_recode. I have useed !!! and haven't succeded .. 

Each input to fct_recode must be a single named string. Problems at positions: 1, 2

batpigandme:
Commenting on a closed issue is pretty low visibility. Before opening a new thread here, let's see if we can get things sorted on https://community.rstudio.com. Do you mind asking it over there? (You might want to read https://www.tidyverse.org/help/ first to maximise your chances of getting a good answer.)

PR #88, opened 2017-04-18
Allow fct_recode to take arguments as list

eclarke:
Hi Hadley, 

### Issue:
When calling `fct_recode`, I would like to be able to specify a list of recodings like so:
```r
recodings = list(x="a", y="b")
fct_recode(f, recodings)
```
However, this gives an error because check_factor_levels converts the list to a list-of-a-list, which breaks the `is_ok` check:
```r
check_recode_levels <- function(...) {
  levels <- list(...)

  is_ok <- function(x) is.character(x) && length(x) == 1
  ok <- vapply(levels, is_ok, logical(1))
  # [... truncated]
```

Right now, to get around this, I have to call do.call on fct_recode like so:
```r
f = do.call(function(...) fct_recode(f, ...), recodings)
```

### This pull request
I've modified `fct_recode` so that it optionally takes a named argument `.levels`. The old behavior is maintained - `fct_recode(f, x='a', y='b')` - but we can also do `fct_recode(f, recodings)` and also mix behaviors: `fct_recode(f, recodings, z='c')`

This should also provide a fix for #84 if the author wraps their argument in a list.

I have written a test for this behavior and modified the two `check_recode_levels` tests to be compatible.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,7 +19,7 @@ Imports:
     tibble,
     magrittr
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 5.0.1.9000
+RoxygenNote: 6.0.1
 Suggests:
     ggplot2,
     testthat,
--- a/R/recode.R
+++ b/R/recode.R
@@ -1,6 +1,8 @@
 #' Change factor levels by hand
 #'
 #' @param f A factor.
+#' @param .levels A list of named character vectors where the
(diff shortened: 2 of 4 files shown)


hadley:
I think we've now picked a standard way of doing this across the tidyverse - use `rlang::dots_list()` so you can use `!!!` to splice a list. Do you want to have a go at that?

PR #96, opened 2017-09-08
Fix proportions in ggplots in v0.1.0 release notes

### LINKED ISSUE #95: viz examples in release 0.1.0 notes are broken
The graphical examples at the bottom of  [release notes for version 0.1.0](http://forcats.tidyverse.org/articles/releases/forcats-0.1.0.html) are broken. I think it's related to a grouped `mutate` call when the call should be ungrouped. 

I think it's a one-line fix, which I will submit a PR for in a minute.

BarkleyBG:
Fixes #95


hadley:
Thanks!

PR #98, opened 2017-09-27
Allow NA in fct_inorder()

### LINKED ISSUE #89: as_factor () Error: `idx` must contain one integer for each level of `f`
column <- c("hello", "dear", "hello", NA, "dear")
column <- forcats::as_factor (column)
Error: `idx` must contain one integer for each level of `f`

I think in this case three options could be considered:
1. convert NA values to a factor with level = "NA";
2. give a more meaningful error message;
3. or just leave NA values as they are, to be consistent with as.factor ()  which handles NA values.

I think the third option is the best, but I do not see what other behavior may  be considered.

krlmlr:
and `as_factor()`. Closes #89.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # forcats 0.2.0.9000
 
+* `as_factor()` and `fct_inorder()` accept NA levels (#98).
+
 # forcats 0.2.0
 
 ## New functions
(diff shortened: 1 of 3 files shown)


jimhester:
The issue this fixes #89, is one of the most upvoted issues in all of the tidyverse packages, so it would be useful to review and merge this to close the issue.

Issue #100, opened 2017-11-29
Integrate fct_other functionality into fct_collapse


jmpsteen:
Thanks for this great package! I'm wondering though whether there would be any possibility of integrating features of `fct_other` into the `fct_collapse` function?

I currently need to recode a factor by collapsing a targeted set of levels (say `c("A", "B", "C", "D", "E")`) into new levels (say `c("AB", "CDE")`) and I want to recode all non-targeted levels (which usually is quite a long set of levels to manually specify, e.g. all other letters of the alphabet) into an `"other"` category. In other words: I'ld like `fct_collapse` to have an additional argument (similar to `fct_other`'s `other_level` argument) that, when specified, automatically identifies all non-targeted levels (i.e. those not specified in any of the vectors, which in `fct_other` would be specified in the `drop` argument) and collapses these into a category with level name specified in this argument.

This feature would be quite helpful, because the only solution I can think of now is quite cumbersome. That is, I have to manually specify all new (collapsed) factor level names (entered in the `fct_collapse` call) in the `keep` argument of `fct_other`, while I assume these could easily be internally identified automatically in the `fct_collapse` function.

holstius:
+1 for this. I often have use for a tabulation where the final element is “(other)”. 

Lexically, the leading parenthesis makes “(other)” follow alphanum levels, so I’d suggest that as a default.

hadley:
Can you please provide a minimal reprex (reproducible example)? The goal of a reprex is to make it as easy as possible for me to recreate your problem so that I can fix it: please [help me help you](https://nypdecider.files.wordpress.com/2014/08/help-me-help-you.gif)! 

If you've never heard of a reprex before, start by reading "[What is a reprex](http://reprex.tidyverse.org/articles/reprex.html)", and follow the advice further down the page. Please make sure your reprex is created with the reprex package as it gives nicely formatted output and avoids a number of common pitfalls.

BirgerNi:
+1 for this.

``` r
suppressMessages(library(tidyverse))

set.seed(123)
data <- sample(LETTERS[1:10], 99, replace = TRUE) %>%
  as_factor()

# What I had to do
data %>% 
  fct_collapse(AB = c("A", "B"), CDE = c("C", "D", "E")) %>%
  fct_other(keep = c("AB", "CDE"))
#>  [1] CDE   Other CDE   Other Other AB    Other Other Other CDE   Other
[...]

PR #101, opened 2017-11-29
Update collapse.R

jmpsteen:
possible solution for issue #100


holstius:
Could fct_collapse coerce non-enumerated values to NA? Then maybe pair it with fct_coalesce? The latter would need to shoulder the burden of adding a new factor level in the case that it’s not already present. 

Alternatively, maybe a special arg “.other” could trail the ellipses in fct_collapse.

hadley:
I'd like to understand this problem a bit better before considering a PR. Let's discuss in the issue.

PR #111, opened 2018-02-11
add '(or character vector)' to function argument descriptions where a…

### LINKED ISSUE #102: Documentation says input is factor, but can be character
Documentation for functions (e.g. `fct_reorder`) say that the first input is a factor. But because of `check_factor`, they can actually be a character instead and will be converted to a factor. I think it would be clearer to surface this, so people aren't needlessly doing `as.factor` before using forcats functions. If you agree, I can make a pull request to update the documentation.

robinsones:
Fixes #102. I went through and verified which functions can accept a character vector. `lvls_reorder` and `lvls_revalue` both can because of `check_factor`, but `lvls_expands` cannot, so I kept the `lvls` documentation as accepting only a factor.

### DIFF
--- a/R/collapse.R
+++ b/R/collapse.R
@@ -1,6 +1,6 @@
 #' Collapse factor levels into manually defined groups
 #'
-#' @param f A factor.
+#' @param f A factor (or character vector).
 #' @param ... A series of named character vectors. The levels in
 #'   each vector will be replaced with the name.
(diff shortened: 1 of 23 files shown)


hadley:
Thanks - I don't see any reason why `lvls_expand()` shouldn't be consistent and use `check_factor()`

Issue #91, opened 2017-06-03
FR: Pass fct_relabel fun parameter to rlang::as_function


alistaire47:
It would be handy to pass the `fun` parameter of `fct_relabel` to `rlang::as_function` so it can accept formula (`~tolower(.x)`) or quosure (`quo(tolower(.x)`) style functions or string function names (`"tolower"`) instead of just raw functions.

A trivial reprex (though realistically regex is the obvious use-case):

``` r
library(tidyverse)
library(forcats)

# Right now this is the only option for an anonymous function:
iris %>% 
    tbl_df() %>% 
    mutate(Species = fct_relabel(Species, 
                                 function(x){substr(x, 1, 3)}))
#> # A tibble: 150 x 5
#>    Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#>           <dbl>       <dbl>        <dbl>       <dbl>  <fctr>
#>  1          5.1         3.5          1.4         0.2     set
#>  2          4.9         3.0          1.4         0.2     set
#>  3          4.7         3.2          1.3         0.2     set
#>  4          4.6         3.1          1.5         0.2     set
#>  5          5.0         3.6          1.4         0.2     set
#>  6          5.4         3.9          1.7         0.4     set
#>  7          4.6         3.4          1.4         0.3     set
#>  8          5.0         3.4          1.5         0.2     set
#>  9          4.4         2.9          1.4         0.2     set
#> 10          4.9         3.1          1.5         0.1     set
#> # ... with 140 more rows

# ...but this syntax would be convenient and consistent with the rest of the tidyverse:
iris %>% 
    tbl_df() %>% 
    mutate(Species = fct_relabel(Species, 
                                 ~substr(.x, 1, 3)))
#> Error in mutate_impl(.data, dots): Evaluation error: Expected function, got formula.

# ...and it would work:
iris %>% 
    tbl_df() %>% 
    mutate(Species = fct_relabel(Species, 
                                 rlang::as_function(~substr(.x, 1, 3))))
#> # A tibble: 150 x 5
#>    Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#>           <dbl>       <dbl>        <dbl>       <dbl>  <fctr>
#>  1          5.1         3.5          1.4         0.2     set
#>  2          4.9         3.0          1.4         0.2     set
#>  3          4.7         3.2          1.3         0.2     set
#>  4          4.6         3.1          1.5         0.2     set
[...]

PR #112, opened 2018-02-12
Use rlang to coerce fct_relabel input to function (#91)

alistaire47:
This PR changes `fct_relabel` per issue #91 so it uses `rlang::as_function` to treat the `fun` parameter, enabling the use of string names of functions, formula functions, and quosure functions.

Notes on changes:

- As discussed in the issue, this introduces an rlang `Imports` dependency.
- I dropped the sanity check on whether `fun` is a function, as `as_function` will always error beforehand anyway. 
  - I changed the accompanying test to reflect the new behavior, but given that `as_function` error messages could change, it may be better to drop the test altogether. For now, I left it with just `"function"` as a regex, which is broad enough that it will probably never break even if errors do change.
- I added a test to make sure it works with something that requires coercion to a function.
- I changed the parameter documentation to reflect the change. It uses the word "coercible", which `devtools::spell_check` doesn't like with my current dictionary, [but it is a word](http://www.dictionary.com/browse/coercible).
- I added an example of the new behavior.
- I updated NEWS.md.

All tests and checks pass on MacOS. 

Let me know if you'd like any reversions or further changes!


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # forcats 0.2.0.9000
 
+# `fct_relabel()` now accepts objects coercible to functions
```
hadley: This should start with `*`, not `#`

NEWS.md
```diff
@@ -1,5 +1,8 @@
 # forcats 0.2.0.9000
 
+# `fct_relabel()` now accepts objects coercible to functions 
+  by `rlang::as_function` (#91)
```
hadley: Can you please add your github user name as described in http://style.tidyverse.org/news.html#during-development ?

DESCRIPTION
```diff
@@ -16,7 +16,8 @@ Depends:
     R (>= 3.1)
 Imports: 
     magrittr,
-    tibble
+    tibble,
```
hadley: These should be in alphabetical order (that should resolve the merge conflict)

R/relabel.R
```diff
@@ -1,9 +1,9 @@
 #' Automatically relabel factor levels, collapse as necessary
 #'
 #' @param f A factor.
-#' @param fun A function that is applied to each level. Must accept one
-#'   character argument and return a character vector of the same length as its
-#'   input.
+#' @param fun A function or something coercible to one by [rlang::as_function()]
```
hadley: I think it's more useful to describe exactly what it can be here, rather than requiring the user to read more technical documentation.

R/relabel.R
```diff
@@ -22,10 +22,14 @@
 #'
 #' rincome2 <- fct_relabel(gss_cat$rincome, convert_income)
 #' fct_count(rincome2)
+#'
+#'
```
hadley: Can you please think through the use of blank lines here? Generally they should be used to group together related code. (This may require tweaking to the entire example block)

R/relabel.R
```diff
@@ -22,10 +22,14 @@
...
+#' fct_count(gss_cat$partyid)
+#'
+#' partyid2 <- fct_relabel(gss_cat$partyid, ~gsub(",", ", ", .x))
```
hadley: I think this example should come first because it's simpler

tests/testthat/test-fct_relabel.R
```diff
@@ -36,3 +36,10 @@ test_that("additional arguments", {
 
   expect_identical(fct_relabel(f1, paste0, "."), factor(paste0(letters, ".")))
 })
+
+test_that("function coercion", {
+  f1 <- factor(letters)
+
+  expect_identical(fct_relabel(f1, ~paste0(.x, ".")),
```
hadley: Can you please use the formatting described at http://style.tidyverse.org/syntax.html#long-lines ?

tests/testthat/test-fct_relabel.R
```diff
@@ -36,3 +36,10 @@ test_that("additional arguments", {
 
   expect_identical(fct_relabel(f1, paste0, "."), factor(paste0(letters, ".")))
 })
+
+test_that("function coercion", {
```
hadley: This should be a sentence

R/relabel.R
alistaire47: I changed the new one to pipes and put the function definition at the start of the second one, so now each example has no blank lines within it, and there's one between them. The `fct_count(gss_cat$rincome)` doesn't _have_ to come after the function definition, but I think it's more readable in terms of whitespace.

tests/testthat/test-fct_relabel.R
alistaire47: The others in this file are concise. It's now a sentence, though not capitalized and with a period. If you want that (or for the others to be changed) let me know.

hadley:
Thanks! 

For future reference, if you include `Fixes #xyz` in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

PR #114, opened 2018-02-15
Rebuild site to deal with blank lines

batpigandme:


batpigandme:
@hadley installed `pandoc 2.1.1`— my work here is done…

PR #116, opened 2018-02-20
📰 Update pkgdown site for 0.3.0.

batpigandme:
* Adds 0.3.0 release info to News.
* Builds other sections with updated pandoc rendering (sections rather than anchors).
* n.b. figures re-rendered, but are exactly the same.


hadley:
Thanks!

PR #120, opened 2018-03-16
Add google analytics to pkgdown site.

batpigandme:
* Also props fixes from pkgdown update! 🎉

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -1,5 +1,5 @@
 template:
-  ganalytics: "UA-115082821"
+  ganalytics: "UA-115082821-1"
   package: tidytemplate
   default_assets: false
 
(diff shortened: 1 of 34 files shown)


batpigandme:
@hadley, I think this one should be good to go with all the pkgdown updates, etc.

_pkgdown.yml
```diff
@@ -1,4 +1,5 @@
 template:
+  ganalytics: "UA-115082821-1"
```
hadley: Does this actually do anything with the tidytemplate template?

batpigandme: Oh, you think it's overridden by the UA in the tidytemplate footer? The way it keeps telling me to check is with this Tag Assistant extension, which doesn't work on local previews…so I honestly can't tell. I was missing the trailing `-1` on the dplyr one deployed, so I don't have a dummy version with which to check.

Issue #126, opened 2018-04-20
Proposal: fct_match with validation


jonocarroll:
(Redesigned from https://github.com/tidyverse/dplyr/issues/3514)

An issue I've faced many times is filtering a `data.frame`/`tibble` based on "levels" (unique values or factor levels) of a column. A common (?) pattern for this is to create a logical vector of presence/absence of the levels compared to a vector of acceptable levels, which `filter` then accepts

```r
filter(iris, Species %in% c("setosa"))
```

The problem with this is that while `filter` and `%in%` are both doing their jobs correctly, together they can introduce an unexpected result

```r
filter(iris, Species %in% c("virginica", "samosa"))
```

One might reasonably expect an error in the attempted filter to levels not present in the data, in the same way that `select` fails loudly if a column is not present

```r
select(iris, Species, Spaceship)
#> Error in FUN(X[[i]], ...) : object 'Spaceship' not found
```

The easiest way to get into this situation is to misspell a level, in which case it will (silently) not be in the filtered result

```r
filter(iris, Species %in% c("selosa", "virginica")) %>% {unique(.$Species)}
#> [1] virginica
#> Levels: setosa versicolor virginica
```

Neither `filter` nor `%in%` is at fault here, but the intent of the code is lost because the implementation is not specific enough: `filter` does not perform "filter to the rows which contain X" but rather "filter to the rows for which some condition is `TRUE`" and delegates that responsibility of identifying those to `%in%`. `%in%` knows nothing of the intent so it responds faithfully with "X is not in this vector".

I propose a new function `fct_match` (and its counterpart `fct_exclude`) which performs validation that the requested levels are indeed contained in the vector prior to generating the logical result of which elements correspond to these levels

```r
fct_match(iris$Species, "selosa", "virginica")
#>  Error: Level(s) not present in factor: "selosa" 
```

and otherwise generates the logical vector, e.g. to be passed to `filter`

```r
fct_match(iris$Species, "setosa", "virginica") %>% table()
#FALSE  TRUE 
#   50   100 

filter(iris, fct_match(Species, "virginica")) %>% head()
#  Sepal.Length Sepal.Width Petal.Length Petal.Width   Species
#1          6.3         3.3          6.0         2.5 virginica
#2          5.8         2.7          5.1         1.9 virginica
#3          7.1         3.0          5.9         2.1 virginica
#4          6.3         2.9          5.6         1.8 virginica
#5          6.5         3.0          5.8         2.2 virginica
[...]

PR #127, opened 2018-04-20
Proposal: fct_match with validation

jonocarroll:
Addresses #126.

Redesign of https://github.com/tidyverse/dplyr/issues/3514, with improvements:

 - distinguishes between `levels(f)` and `unique(f)` using `allow_missing`.
 - full test coverage.
 - cleaner API.


hadley:
Are you still interested in this? I think the basic approach is ok, but there are quite a few style issues..

jonocarroll:
Absolutely. I'm happy to put in some effort to bring it into line with acceptable style.

DESCRIPTION
```diff
@@ -25,4 +25,4 @@ Suggests:
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 6.0.1
+RoxygenNote: 6.0.1.9000
```
hadley: Can you please rebuild with CRAN roxygen? (You will probably need to merge recent changes from master to avoid spurious diffs)

R/match.R
```diff
@@ -0,0 +1,101 @@
+#' Test for presence of certain groups/levels in a vector
+#'
+#' A validated version of `%in%` intended for use with logical subsetting.
+#' Performs matching between a factor and a set of levels to pose the question
+#' "which of the elements of a factor match these levels?"
+#'
+#' @md
+#' @rdname fct_match
+#' @param f A factor (or character vector).
+#' @param inverse (logical) if `TRUE`, perform a "not match" operation.
```
hadley: This argument and `fct_exclude()` do not seem compelling to me as they're neither more expression or more succinct than `!fct_match()`

R/match.R
```diff
@@ -0,0 +1,101 @@
...
+#' @param allow_missing (logical) if `TRUE`, allow matching to empty levels in
+#'   f
+#' @param ... A series of characters for which the test "are these levels in
```
hadley: Are you sure this should be `...` and not a single vector? Using `...` saves typing but it makes it feel a bit weird to me.

R/match.R
```diff
@@ -0,0 +1,101 @@
...
+#'   `f`?" should be performed.
+#'
+#' @return a logical vector determined by whether or not the levels in `...` are
```
hadley: Should start with capital letter. But I think you can just say `A logical vector`; the rest is implied by the other documentation.

R/match.R
```diff
@@ -0,0 +1,101 @@
...
+#'   present in `f`
+#'
+#' @details Unlike `%in%`, which performs a basic yes/no test, this function
```
hadley: Can you please place after the description block?

R/match.R
```diff
@@ -0,0 +1,101 @@
...
+#'
+#' A validated version of `%in%` intended for use with logical subsetting.
+#' Performs matching between a factor and a set of levels to pose the question
+#' "which of the elements of a factor match these levels?"
+#'
+#' @md
+#' @rdname fct_match
+#' @param f A factor (or character vector).
+#' @param inverse (logical) if `TRUE`, perform a "not match" operation.
+#' @param allow_missing (logical) if `TRUE`, allow matching to empty levels in
```
hadley: I'd prefer to start simple, and omit this argument for now.

R/match.R
```diff
@@ -0,0 +1,101 @@
...
+#'   not present, then an error is thrown rather than producing  a silent
+#'   `FALSE`.
+#'
+#'   Since a distinction can be made between `levels(f)` and `unique(f)` (the
+#'   latter excludes empty levels) the argument `allow_missing` can be
+#'   specified.
+#'
+#'   `fct_exclude` is a shortcut to `fct_match(inverse = TRUE)`.
+#'
+#'   This is intended to replace the common pattern
```
hadley: I think this a nice summary of the intent and hence should appear in the function description.

R/match.R
```diff
@@ -0,0 +1,101 @@
...
+#'   `filter(d, l %in% c("x", "y"))`
+#'
+#'   with the validated
+#'
+#'   `filter(d, fct_match(l, c("x", "y")))`
+#'
+#'   where an error will be generated if `"x"` or `"y"` are not levels in `d$l`.
+#'
+#' @export
+#' @importFrom rlang dots_list
```
hadley: Please drop this import

R/match.R
```diff
@@ -0,0 +1,101 @@
...
+#'
+#' @examples
+#' ## TRUE if present
```
hadley: Please use a single `#` for comments, and I don't think this level of commenting is useful.

R/match.R
```diff
@@ -0,0 +1,101 @@
...
+  fct_match(f, ..., inverse = TRUE, allow_missing = allow_missing)
+}
+
+
+#' Validate levels are present at all in a factor
+#' @keywords internal
+#' @noRd
+validate_level_inputs <- function(factor, check) {
+  levels_present <- check %in% factor
+  if (!all(levels_present)) {
```
hadley: An early return would make this more readable:

```R
if (all(levels_present))  {
  return(TRUE)
}
```

R/match.R
```diff
@@ -0,0 +1,101 @@
...
+    stop(
+      paste0(
+        "Level(s) not present in factor: \"",
+        toString(check[which(!levels_present)]),
```
hadley: Can you please use `encodeString()` here? (Since I actually know what it does because we use it elsewhere in the tidyverse)

tests/testthat/test-fct_match.R
```diff
@@ -0,0 +1,25 @@
+context("test-fct_match.R")
+
+test_that("correct matching as expected", {
+  f <- factor(c("a", "b", "c d"), levels = c("a", "b", "c d", "e"))
+
+  expect_equal(fct_match(f, "a", "b"), c(TRUE, TRUE, FALSE))
+  expect_error(fct_match(f, "a", "x"), regexp = 'not present.*"x"')
```
hadley: Please don't name the second argument here.

DESCRIPTION
jonocarroll: Upgraded to 6.1.1

R/match.R
jonocarroll: Done.

R/match.R
jonocarroll: Done.

R/match.R
jonocarroll: Done.

R/match.R
jonocarroll: Done. I was previously returning `NULL` (as the output is not used) but `TRUE` makes more sense.

R/match.R
jonocarroll: Done. I was not aware of that one, but I use `toString` to convert vectors to comma-separated strings. I've seen it in base code, I'm sure. `encodeString` seems to additionally do some escaping.

tests/testthat/test-fct_match.R
jonocarroll: Done. I assumed that would be helpful to distinguish it from the tested expression, but no worries if not.

R/match.R
jonocarroll: Counterpoint: using this within a pipe, e.g.

```r
gss_cat$marital %>% 
  fct_exclude("No answer") %>% 
  table()
```

vs

```r
gss_cat$marital %>% 
  fct_match("No answer") %>% 
  `!` %>% 
  table()
```

jonocarroll:
Tests and checks all passing. I'm happy to make more changes if you have suggestions.

R/match.R
jonocarroll: Which should be the default behaviour, then? `levels()` or `unique()`?

hadley:
Thanks! I simplified the code and docs a bit, and also handled NAs to match `%in%`.

Issue #148, opened 2018-10-09
Feature Request: Readme.md examples of usage


hack-r:
I hate to complain, but it strikes me as very odd and unfortunate that the Read Me doesn't actually include any example usage or comments on `forcats` functions... it just talks about factors in general, to the extent you'd probably already be familiar with them when you're researching this library.

Explaining some concrete use cases and key functions of `forcats` would be a major improvement IMHO

batpigandme:
@robinsones,
Would you still be interested in doing a PR for this or non-README forcats content? (I remember we had a discussion about this a while back, but I can't for the life of me find it on GitHub 😞).

robinsones:
Yes, I actually was just working on it! It's getting a bit long so I wonder if a vignette is better ... but @batpigandme why don't I do a pull request with what I have and you can see?

batpigandme:
Sounds good! No rush.

PR #149, opened 2018-10-11
Adding examples to README

robinsones:
addresses #148


batpigandme:
Thanks, Emily! Just started reading, and it's really good— I'm thinking that from the ~conversational tone (which I like, btw) it's probably more vignette than README, but either way, this is excellent material. (actual review notes pending)

robinsones:
I'd be up for making it a vignette and writing a shorter summary for a README. Something like: 

The goal of the forcats package is to provide a suite of useful tools that solve common problems with factors. For example: 
- Reordering a graph (`fct_reorder`, `fct_infreq`, `fct_rev`)
- Collapsing levels together manually (`fct_collapse`)
- Changing some levels to "other" either manually (`fct_other`) or based on frequency (`fct_lump`)
- Changing the order of the factor (`fct_relevel`)

Please see the vignette for details.

batpigandme:
Awesome — I'll finish the review (since it's just copy-edit stuff; sorry, got sidetracked), and then we can figure out what goes where, etc. 🎉

README.Rmd
```diff
@@ -46,28 +46,113 @@ forcats is now part of the core tidyverse, so you do not need to load it explici
...
-Factors are used to describe categorical variables with a fixed and known set of __levels__. You can create factors with the base `factor()` or [`readr::parse_factor()`](http://readr.tidyverse.org/reference/parse_factor.html):
+Let's try answering the question "what are the most common homeworld of starwars characters?" Let's start off by making a barplot
+
+```{r initial_plot}
+ggplot(starwars, aes(x = hair_color)) + 
+  geom_bar() + 
+  coord_flip()
+```
+
+That's okay, but it would be more helpful the graph was ordered by count. This is a case of an **unordered** categorical variable where we want it ordered by its frequency. We can use the function `fct_infreq`:
```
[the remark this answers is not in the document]
batpigandme: Also "house style" thing is to put the parens after functions so here it would be `fct_infreq()`

README.Rmd
```diff
@@ -46,28 +46,113 @@ forcats is now part of the core tidyverse, so you do not need to load it explici
...
+
+What if we want to do a similar plot for skin color? 
+
+```{r fct_infreq_skin}
+ggplot(starwars, aes(x = fct_infreq(skin_color))) + 
+  geom_bar() + 
+  coord_flip()
+```
+
+Here we have a lot of different skin colors - let's reduce it to only be the top 5. We can use `fct_lump` to "lump" all the infrequent colors into one factor, "other." The argument `n` is the number of levels we want to keep.
```
batpigandme: `fct_lump()`

README.Rmd
```diff
@@ -46,28 +46,113 @@ forcats is now part of the core tidyverse, so you do not need to load it explici
...
+
+```{r fct_lump_prop}
+ggplot(starwars, aes(x = fct_lump(skin_color, prop = .1))) + 
+  geom_bar() + 
+  coord_flip()
+```
+
+Only light and fair remain; everything else is other. 
+
+What if we wanted to see if the average mass differed by eye color? We'll only look at the 6 most popular eye colors and remove NAs.
```
batpigandme: I think we do backticks around `NA`s as well. 👍

README.Rmd
```diff
@@ -46,28 +46,113 @@ forcats is now part of the core tidyverse, so you do not need to load it explici
...
+```{r fct_lump_mean}
+starwars %>%
+  mutate(eye_color = fct_lump(eye_color, n = 6)) %>%
+  group_by(eye_color) %>%
+  summarise(mean_mass = mean(mass, na.rm = TRUE)) %>%
+  ggplot(aes(x = eye_color, y = mean_mass)) + 
+  geom_col()
+```
+
+It looks like people (or at least one person) with orange eyes is definitely heavier! This graph would look nice though if it was ordered by `mean_mass`. We can do this with `fct_reorder`, which reorders one variable by another.
```
batpigandme: * I think it's people…are, since "one person" is in parens.
* parens for `fct_reorder()` ( just adding another set of eyes for these, since they're easy to miss with args, etc. in the mix)!

README.Rmd
```diff
@@ -46,28 +46,113 @@ forcats is now part of the core tidyverse, so you do not need to load it explici
...
+
+Let's switch to using another dataset, `gss_cat`, the general social survey. What is the income distribution among the respondents? 
+
+```{r initial_plot2}
+ggplot(gss_cat, aes(x = rincome)) +
+   geom_bar() + 
+   coord_flip()
+```
+
+Notice that the income levels are in the correct order - they go from `"Lt $1000"` to `$25,000 or more`. This is not a coicidence. When you're working with ordinal data, where there is an order, you can have an ordered factor. You can examine them with the base function `levels`, which prints them in order:
```
batpigandme: * Great. I really like how you introduce this. 
* Parens for `levels()`

README.Rmd
```diff
@@ -46,28 +46,113 @@ forcats is now part of the core tidyverse, so you do not need to load it explici
...
-x1 <- c("Dec", "Apr", "Jan", "Mar")
-month_levels <- c(
-  "Jan", "Feb", "Mar", "Apr", "May", "Jun", 
-  "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"
-)
+levels(gss_cat$rincome)
+```
 
-factor(x1, month_levels)
+But what if your factor came in the wrong order? Let's simulate that by reordering the levels of `rincome` randomly with `fct_shuffle` and then plotting it again.
```
batpigandme: Paren `fct_shuffle()`

README.Rmd
```diff
@@ -46,28 +46,113 @@ forcats is now part of the core tidyverse, so you do not need to load it explici
...
 
-parse_factor(x1, month_levels)
+```{r fct_shuffle}
+ggplot(gss_cat, aes(x = fct_shuffle(rincome))) +
+   geom_bar() + 
+   coord_flip()
 ```
 
-The advantage of `parse_factor()` is that it will generate a warning if values of `x` are not valid levels:
+How can we fix this and put it in the right order?
```
batpigandme: Maybe have a subheader in here about reordering for navigability?

README.Rmd
```diff
@@ -46,28 +46,113 @@ forcats is now part of the core tidyverse, so you do not need to load it explici
...
+
+We can use the function `fct_relevel` when we need to manually reorder our factor levels. In addition to the factor, you give it a character vector of level names and specify where you want to move them. It defaults to moving them to the front, but you can move them after another level with the argument `after`. If you want to move it to the end, you set `after` equal to `Inf`.
```
batpigandme: * `fct_relevel()`
* Comma after "level names" before "and"

README.Rmd
robinsones: That sounds helpful. What do you think of adding a couple headers, so we don't have the large "getting started" chunk? Could do: 

Ordering by frequency (instead of getting started)

Combining levels
 
Combining by another variable

Manually reordering

robinsones:
Thanks! Think I got everything in the newest push

hadley:
I think this would be better in a vignette (called forcats) with a brief summary in the README. Want to finish it off on tidyverse dev day?

robinsones:
@batpigandme could you add the tidyverse-dev-day tag?

batpigandme:
@robinsones done!

robinsones:
great, I've got it!

hadley:
Superseded by #158

PR #152, opened 2018-12-10
fct_relabel accepts character input

billdenney:
This PR makes `fct_relabel()` have parallel functionality with respect to character input as other `fct_*()` functions.


hadley:
Thanks!

PR #154, opened 2019-01-04
Allow `fct_reorder` to work with non-numeric functions (Fix #147)

### LINKED ISSUE #147: Feature Request: Allow fct_reorder to Work with Non-Numeric Results of .fun
This is the same issue as #103 with a different request.

It seems as though `fct_reorder()` and `fct_reorder2()` should only require that the results of the function are sortable not that they are numeric.  Lexicographical order for the below is consistent and well-defined, but a simple numerical mapping can be more complex when the input isn't as simple as the below.

My feature request is:  Can `fct_reorder()` and `fct_reorder2()` please allow non-numeric values?  It seems like it could be as simple as running `xtfrm()` on the results of `.fun`.

Related to this, can the documentation be updated to indicate what the return value of `.fun()` must be?

``` r
library(forcats)
fct_reorder(.f=LETTERS[1:3], .x=LETTERS[4:6], .fun=min)
#> Error: `fun` must return a single number per group
```

Created on 2018-09-11 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

billdenney:


billdenney:
This fixes #147

hadley:
I think it's adequate to just relax the constraint on the return type of `tapply()`, which was just as easy for me to do. Thanks for helping me think about this.

PR #156, opened 2019-01-08
Add `fct_sort` function

### LINKED ISSUE #117: Feature request: reorder factor levels using a user-supplied sorting function
Hi,
I was wondering whether you might consider adding a function to `forcats` that sets up factor levels according to a numerically-aware string-sorting method like gtools::mixedsort. 
There's a number of use cases: eg, when you end up with genomes ordered like "chr1", "chr10", "chr11" ... "chr2" rather than "chr1", "chr2", ..., in a plot if you don't hack your factors correctly

eg, 
```
fct_alphameric/fct_alnum/fct_mixedsort <- function(x, ...) {
    factor(x, levels = gtools::mixedsort(unique(x), ...))
    }
```

russHyde:
`fct_sort` takes a factor or character-vector (implictly converted to factor) and reorders the `levels` of that factor using a user-specified function `.fun`.

Code added to `R/sort.R`, unit tests into `tests/testthat/test-fct_sort.R`.

An example showing how to use `fct_sort` to sort number-containing character `levels` by the contained number.

Ensures that the values in the factor levels are unchanged by the sorting function, and uses `...` to pass further arguments for the sorting function.


hadley:
Fixes #117

R/sort.R
```diff
@@ -0,0 +1,38 @@
...
+#'  }
+#'  fct_sort(chr_fac, sort_numeric)
+#'
+fct_sort <- function(.f, .fun, ...) {
+  f <- check_factor(.f)
+  .fun <- rlang::as_function(.fun)
+
+  old_levels <- levels(f)
+  new_levels <- .fun(old_levels, ...)
+  stopifnot(
```
hadley: Would you mind making this a slightly friendlier function using the style guide at http://style.tidyverse.org/error-messages.html ?

R/sort.R
```diff
@@ -0,0 +1,38 @@
+#' Automatically sort factor levels according to a user-defined function
```
hadley: Maybe remove "Automatically"?

R/sort.R
```diff
@@ -0,0 +1,38 @@
...
+#' @export
+#' @examples
+#' chromosomes <- c("chr2", "chr1", "chr10")
+#' chr_fac <- factor(chromosomes, levels = chromosomes)
+#' chr_fac
+#'
+#' # naive alphanumeric sorting "1" < "10" < "2"
+#' fct_sort(chr_fac, sort)
+#'
+#'  # number-based alphanumeric sorting "1" < "2" < "10"
```
hadley: I think this is maybe a bit complicated for an example? Maybe just do something with alphabetical sorting? And maybe `sample()` to show random reordering?

jennybc:
The name `fct_sort()` feels potentially misleading to me, as it's not clear if it acts on the factor (what I would assume) or its levels (what actually happens).

Worth contemplating: could this be achieved this by making `fct_relevel()` more capable? I.e. make it work more like `purrr::set_names()`, which can accept a vector of names, individual names, or a function or formula.

If that's a no-go, I still think it's worth trying to think of a better name.

huftis:
> The name `fct_sort()` feels potentially misleading to me, as it's not clear if it acts on the factor (what I would assume) or its levels (what actually happens).

The same is true for other `forcats` functions, like `fct_rev()` and `fct_shuffle()`.

But perhaps `fct_sort_levels()` would be work? It doesn’t really take any longer time to write if you have auto-completion enabled.

russHyde:
Totally agree with @jennybc  and @huftis  that a better name should be applied. Happy to rewrite as `fct_sort_levels`. 

I didn't consider modifying the `fct_relevel` formals, but that is a good proposal and would be possible, eg, as `fct_relevel(.f, ..., after = 0L, .fun)`

huftis:
> I didn't consider modifying the `fct_relevel` formals, but that is a good proposal and would be possible, eg, as `fct_relevel(.f, ..., after = 0L, .fun)`

I would prefer having a separate `fct_sort_levels()` function. IMHO, having the sorting function as the second argument is much nicer. It’s faster and feels cleaner to write `fct_sort_levels(x, mysort)` than `fct_relevel(x, .fun = mysort)`.

hadley:
I think @jennybc was suggesting that `fct_relevel(x, mysort)` could be made to work.

R/sort.R
russHyde: Added more informative messages to indicate: 

- when .fun returns with something other than a vector

- when the sorted-levels are of different length from the input-levels

- when the sorted-levels contains at least one level that is absent from the input-levels

R/sort.R
russHyde: The description now starts `#' Sort factor levels ...`

R/sort.R
russHyde: I replaced the example. The new examples includes

- alphabetical-sort,
- alphabetical-decreasing-sort (equiv to fct_rev)
- alphabetical sort with an out-of-order baseline level (equiv to fct_relevel)
- sampling from the levels

russHyde:
I believe I have made all changes suggested by @hadley and @huftis .
The function has been renamed `fct_sort_levels`.

R/sort_levels.R
```diff
@@ -0,0 +1,67 @@
+#' Sort factor levels according to a user-defined function
+#'
+#' @param .f A factor (or character vector).
+#' @param .fun A function that will sort or permute the existing factor levels.
```
hadley:
```suggestion
#' @param .fun A function that will permute the existing factor levels.
```

R/sort_levels.R
```diff
@@ -0,0 +1,67 @@
...
+#'   It must accept one character argument and return a character argument of
+#'   the same length as it's input.
```
hadley:
```suggestion
#'   the same length as its input.
```

R/sort_levels.R
```diff
@@ -0,0 +1,67 @@
...
+#' @examples
+#' medieval_experiment <- factor(
+#'   c("Control", "Comfrey", "Cropleek", "Bishopwort")
+#' )
+#' medieval_experiment
+#'
+#' # Default (alphabetical) level-sorting:
+#' fct_sort_levels(medieval_experiment, sort)
+#'
+#' # Reversed ordering (equivalent to `fct_rev`):
```
hadley: It's not in general equivalent to `fct_rev()`, so I think it'd be better to just remove that comparison

R/sort_levels.R
```diff
@@ -0,0 +1,67 @@
...
+#' fct_sort_levels(medieval_experiment, sort, decreasing = TRUE)
+#'
+#' # Level-sorting with "Control" as the first level
```
hadley: And this seems inferior to `fct_relevel()` so I'd also drop it.

R/sort_levels.R
```diff
@@ -0,0 +1,67 @@
...
+#' # (equivalent: `fct_relevel(medieval_experiment, "Control")`)
+#' fct_sort_levels(
+#'   medieval_experiment,
+#'   function(x) c("Control", sort(setdiff(x, "Control"))))
+#'
+#' # Randomised sorting of the levels:
```
hadley:
```suggestion
#' # Randomly permute the levels:
```

R/sort_levels.R
```diff
@@ -0,0 +1,67 @@
...
+
+  old_levels <- levels(f)
+  new_levels <- .fun(old_levels, ...)
+  if (!is.vector(new_levels)) {
+    stop(
+      paste("The result of `.fun(levels(.f))` must be a vector"),
+      call. = FALSE
+    )
+  }
+  if (length(old_levels) != length(new_levels)) {
```
hadley: This feels a little overwrought to me. Doesn't `fct_relevel()` already give error messages?

russHyde: I don't think it provdes an error in this setting - if passed a (vector) proper subset of a factor's levels, `fct_relevel` just pulls the subset to the start of the factor's levels:
```
fct_relevel(factor(letters[1:3]), c("b", "a"))
[1] a b c
Levels: b a c
```

(I thought the error messages might be overkill, but was trying to follow the examples in your style guide)

R/sort_levels.R
russHyde: @hadley  : Given the emphasis on permutation in your edited version of the docs, would `fct_permute_levels` be a more appropriate name for the function?

hadley:
After thinking about @jennybc's comment more, I think this is best implemented as a special case of `fct_relevel()`. Your PR was very useful to help us think through the API and implementation — thanks!

russHyde:
thanks

PR #159, opened 2019-01-19
add fct_recode example. fixes #140

### LINKED ISSUE #140: `fct_relabel` taking a named vector as its anonymous function
I was working with data where I didn't want to type out each level by hand in a single block with `fct_recode`, so I assumed there had to be a way to pass a named vector to rename levels. That did not work, with the informative message 

> `Evaluation error: Each input to fct_recode must be a single named string. Problems at positions: 1.`

While it wasn't immediately obvious to me why I would guess to try `fct_relabel`, I did, and that works fine:

```r
library(dplyr)
x <- letters[1:16]
names(x) <- levels(gss_cat$relig)

gss_cat %>% mutate(relig = fct_recode(relig, x))    # does not work
gss_cat %>% mutate(relig = fct_relabel(relig, ~x))  # works
```
Which replaces "Protestant" with "o" and "Orthodox-christian" with "f" and so on. (I realize this toy example could have used `fct_anon`, but my real use case was the reverse operation).

Is this intended behavior? Is this supported behavior? I didn't see it documented, but I'd be glad to add an example to the docs if so.

Ryo-N7:
added two examples for passing named vectors to fct_recode to rename levels using !!! to unquote

### DIFF
--- a/R/recode.R
+++ b/R/recode.R
@@ -16,14 +16,10 @@
 #' # If you name the level NULL it will be removed
 #' fct_recode(x, NULL = "apple", fruit = "banana")
 #'
-#' # When passing a named vector to rename levels use !!! to unquote
+#' # When passing a named vector to rename levels use !!! to splice
 #' x <- factor(c("apple", "bear", "banana", "dear"))
 #' levels <- c(fruit = "apple", fruit = "banana")
 #' fct_recode(x, !!!levels)
-#'
-#' # The above but within a dataframe
-#' x_df <- x %>% as_data_frame()
-#' x_df %>% mutate(value = fct_recode(value, !!!levels))
 fct_recode <- function(.f, ...) {
   f <- check_factor(.f)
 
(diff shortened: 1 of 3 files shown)


R/recode.R
```diff
@@ -15,6 +15,15 @@
 #'
 #' # If you name the level NULL it will be removed
 #' fct_recode(x, NULL = "apple", fruit = "banana")
+#'
+#' # When passing a named vector to rename levels use !!! to unquote
```
hadley: I think it's slightly better to say "splice" rather than "unquote"

R/recode.R
```diff
@@ -15,6 +15,15 @@
...
+#' x <- factor(c("apple", "bear", "banana", "dear"))
+#' levels <- c(fruit = "apple", fruit = "banana")
+#' fct_recode(x, !!!levels)
+#'
+#' # The above but within a dataframe
```
hadley: forcats doesn't use on dplyr, so unfortunately we can't use it in tests.

hadley:
Thank you!

PR #160, opened 2019-01-19
Add group_other option to fct_collapse, to allow for optional recoding of all non-targeted levels into an "Other" category

### LINKED ISSUE #100: Integrate fct_other functionality into fct_collapse

AmeliaMN:
Fixes #100

### DIFF
--- a/R/collapse.R
+++ b/R/collapse.R
@@ -3,7 +3,7 @@
 #' @param .f A factor (or character vector).
 #' @param ... A series of named character vectors. The levels in
 #'   each vector will be replaced with the name.
-#' @param group_other Replace all levels not named in `...` to "Other"?
+#' @param group_other Replace all levels not named in `...` with "Other"?
 #' @export
 #' @examples
 #' fct_count(gss_cat$partyid)
--- a/R/collapse.R
+++ b/R/collapse.R
@@ -3,6 +3,7 @@
 #' @param .f A factor (or character vector).
 #' @param ... A series of named character vectors. The levels in
 #'   each vector will be replaced with the name.
+#' @param group_other Replace all levels not named in `...` to "Other"?
 #' @export
 #' @examples
 #' fct_count(gss_cat$partyid)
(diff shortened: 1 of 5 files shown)


hadley:
Can you please give this PR an informative title, and then include `Fixes #100` in the description?

AmeliaMN:
Done!

R/collapse.R
```diff
@@ -15,10 +16,16 @@
 #'   dem = c("Not str democrat", "Strong democrat")
 #' )
 #' fct_count(partyid2)
-fct_collapse <- function(.f, ...) {
+fct_collapse <- function(.f, ..., group_other = FALSE) {
   new <- rlang::dots_list(...)
-
   levs <- as.list(unlist(new, use.names = FALSE))
+  if(group_other){
```
hadley: Missing space before `(`

tests/testthat/test-collapse.R
```diff
@@ -21,3 +21,17 @@ test_that("can collapse missing levels", {
 
   expect_equal(f2, factor(c("x", "y")))
 })
+
+test_that("can collapse un-named levels to Other", {
+  f1 <- factor(letters[1:3])
+  f2 <- fct_collapse(f1, xy = c("a", "b"), group_other=TRUE)
```
hadley: Can you please put spaces around `=`?

hadley:
Looks good!

Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thank you!!

PR #164, opened 2019-01-19
Modify forecast readme + add COC

### LINKED ISSUE #148: Feature Request: Readme.md examples of usage

robinsones:
The README now has a shorter introduction, shares a few examples, removes the section of making factors, and adds a "getting help" section. I also added the tidyverse code of conduct (`usethis::use_tidy_coc()`) after checking with @batpigandme; I put it in the same branch since it modified the README too. This closes #148.


README.Rmd
```diff
@@ -8,8 +8,9 @@ output: github_document
 knitr::opts_chunk$set(
   collapse = TRUE,
   comment = "#>",
-  fig.path = "README-"
+  fig.path = "man/figures/README-"
 )
+library(tidyverse)
```
hadley: Like for other PR you'll need to load specific packages (since we can't take tidyverse as even an optional dependency since it's so heavy)

PR #158, opened 2019-01-19
Add forcats vignette

robinsones:
partially solves #149, making another branch with README changes.

PR #165, opened 2019-01-19
Add forcats vignette

robinsones:
Changed issue with pull request #158 by adding dplyr to suggests, so vignette will build


hadley:
Thanks!

PR #168, opened 2019-01-19
Keeping the same label in fct_lump() if no lumping occurs (Fix #130)

### LINKED ISSUE #130: fct_lump() shouldn't relabel if no lumping occurs
If in a vector, only one value appears less than the 'prop' times, fct_lump creates a new level called 'other'. Ideally, it should keep the original level name as only one level was affected. 

Example:
nRows <- 500
vec <- as.factor(c(rep("X",0.32*nRows),rep("Y",0.08*nRows), rep("Z",0.4*nRows), rep('W', 0.2*nRows)))
rebinned_vec <- fct_lump(vec, prop = 0.1)

prop.table(table(rebinned_vec)) gives the following output:
    W     X     Z Other 
 0.20  0.32  0.40  0.08

In the above code, only the level 'Y' should be affected as it has less than 10% share. But since this is the only level affected, isn't it expected that fct_lump will leave the level 'Y' as it is rather than creating the 'other' level?

zhiiiyang:
This fixes #130


NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 * `fct_count()` gains a parameter to compute the fraction of marginal table (@zhiiiyang, #146). 
 
+* `fct_lump()` now does not change the label if no lummping occurs (@zhiiiyang, #130).
```
hadley:
```suggestion
* `fct_lump()` now does not change the label if no lumping occurs (@zhiiiyang, #130). 
```

tests/testthat/test-fct_lump.R
```diff
@@ -12,7 +12,7 @@ test_that("positive values keeps most commmon", {
 
 test_that("ties are respected", {
```
hadley: I think these need to be updated so that they continue to make "other" levels

R/lump.R
```diff
@@ -74,14 +74,22 @@ fct_lump <- function(f, n, prop, w = NULL, other_level = "Other",
       rank <- rank(-count, ties = ties.method)
     }
 
-    new_levels <- ifelse(rank <= n, levels, other_level)
+    if(sum(!(rank <= n)) == 1){
```
hadley: I can't follow the logic here. Could you simplify somehow? I wonder if it might make sense to do an early return of `f`?

NEWS.md
zhiiiyang: Oops, thanks!

hadley:
Thanks!

PR #170, opened 2019-01-19
Numeric as factor

### LINKED ISSUE #145: no applicable method for 'as_factor' applied to an object of class "c('double', 'numeric')"
I tried to categorize a numerical vector by using as_factor()

`as_factor(c(1, 2, 3))`

The error hapeend. 

> Error in UseMethod("as_factor") : no applicable method for 'as_factor' applied to an object of class "c('double', 'numeric')"

But it works when comes to as.factor()

`as.factor(c(1, 2, 3))`
`[1] 1 2 3`
`Levels: 1 2 3`

kbodwin:
fixes #145

`as_factor` now has a numeric method. By default, orders factors in numeric order, unlike the other methods which default to order of appearance.

This method depends on `fct_sort_num()` which reorders labels in numeric order, if possible.  Note that this is not equivalent to sorting the levels, which defaults to alphabetical order.  For consistency, I also added `fct_sort_alpha()` which reorders labels in English alphabetical order.

```
x <- factor(as.character(c(3.2, 1.1, 11, 22, 2.2, 5)))

levels(x)
#>  [1] "1.1" "11"  "2.2" "22"  "3.2" "5"  

fct_sort_num(x)
#> [1] 3.2 1.1 11  22  2.2 5  
#> Levels: 1.1 2.2 3.2 5 11 22

fct_sort_alpha(x)
#> [1] 3.2 1.1 11  22  2.2 5  
#> Levels: 1.1 11 2.2 22 3.2 5

factor(x, levels = sort(levels(x)))
#> [1] 3.2 1.1 11  22  2.2 5  
#> Levels: 1.1 11 2.2 22 3.2 5

```


kbodwin:
Done, thank you for the tips!

I also renamed `fct_sort_num` to `fct_inseq` to be more consistent with the related functions `fct_inorder` and `fct_infreq`.

R/utils.R
```diff
@@ -16,8 +16,10 @@ check_factor <- function(f) {
     factor(f)
   } else if (is.factor(f)) {
     f
+  } else if (is.numeric(f)) {
```
hadley: This seems a little over broad to me - I think it's ok to permit explicit coercion, but not to do it automatically (as that makes it easier for mistakes to slip through)

R/as_factor.R
```diff
@@ -32,3 +35,18 @@ as_factor.character <- function(x, ...) {
     label = attr(x, "label", exact = TRUE)
   )
 }
+
+#' @rdname as_factor
+#' @export
+as_factor.numeric <- function(x, ...) {
+
+  x <- as.character(x)
```
hadley: I think this would be simpler as `factor(x, levels = sort(unique(x))`?

R/reorder.R
```diff
@@ -97,3 +103,20 @@ fct_infreq <- function(f, ordered = NA) {
...
+
+#' @export
+#' @rdname fct_inorder
+fct_inseq <- function(f, desc = FALSE) {
+  f <- check_factor(f)
+
+  num_levels <- suppressWarnings(as.numeric(levels(f)))
+  new_levels <- sort(num_levels, decreasing = desc)
+
+  if(length(new_levels) == 0){
```
hadley: Missing space after `if`

R/reorder.R
```diff
@@ -97,3 +103,20 @@ fct_infreq <- function(f, ordered = NA) {
 
   lvls_reorder(f, order(table(f), decreasing = TRUE), ordered = ordered)
 }
+
+
+#' @export
+#' @rdname fct_inorder
+fct_inseq <- function(f, desc = FALSE) {
```
hadley: Should this have `ordered = NA` to match the other functions in this family?

This is also the only function to use `desc`, which is surprising. Maybe remove for now?

NEWS.md
```diff
@@ -1,6 +1,8 @@
 # forcats 0.3.0.9000
 
 * `fct_relabel()` now accepts character input.
+* `as_factor` now has a numeric method. By default, orders factors in numeric order, unlike the other methods which default to order of appearance. (#145, @kbodwin)
+* `fct_sort_num()` reorders labels in numeric order, if possible. (#145, @kbodwin)
```
hadley: I think these needs an update

R/as_factor.R
kbodwin: You're right; the character step was left over from when I wanted labels in order of appearance.

Now that you mention it - `factor(x)` is good enough, levels are in numeric order by default.

R/utils.R
kbodwin: I'm happy to make this change, of course, but for learning purposes, can I ask about the philosophy here?

If someone passes a character vector to `fct_inorder()`, `fct_count()`, or similar, we assume they intended to factor that vector, and we coerce it in `check_factor()`.

If someone passes a numeric vector for `fct_inorder()`, why should we assume this was a mistake rather than auto-coercing?

hadley: In an ideal world, I don't think `fct_inorder()` etc would accept either character or factors, but because they're used almost interchangeably in R, I think it would be too frustrating if these functions didn't accept either. But numeric vectors are quite different from factors, so I think it makes sense to require the user to explicit coerce.  It's basically a tradeoff: it makes life a little bit harder sometimes in return for getting clearer error messages when you've actually made a mistake.

hadley:
Thanks @kbodwin!

PR #176, opened 2019-02-18
fix for issue whereby fct_collapse mislabels collapsed factors, where… (Fix #172)

### LINKED ISSUE #172: fct_collapse applying collapsed factors in wrong order when group_other = TRUE
I'm having an issue which is self explanatory from the title. Reprex below. As you can see, the classification produced by `fct_collapse` is incorrect.

As requested, a simpler set of examples is below.

## Example 1 - Collapsing _Character_ Variables
The `fct_collapse` function wrongly classifies _raspberry_ as a vegetable, rather than broccoli.
```r
library(forcats)
library(dplyr)
library(tibble)

df <-
  tibble(item = c("apple", "grape", "banana", "broccoli", "raspberry"))

df %>% 
  mutate(category = fct_collapse(item,
                                 fruit = c("apple", "grape", "banana", "raspberry"),
                                 vegetables = "broccoli", group_other = TRUE))
#> # A tibble: 5 x 2
#>   item      category  
#>   <chr>     <fct>     
#> 1 apple     fruit     
#> 2 grape     fruit     
#> 3 banana    fruit     
#> 4 broccoli  fruit     
#> 5 raspberry vegetables
```
<details>

## Example 2 - Collapsing _Factor_Variables
For clarity, and contrary to my previous explanation, this bug affects both character and factor variables (since `fct_collapse` converts the `.f` argument to a factor if it is a character variable):
```r
df <-
  tibble(item = factor(c("apple", "grape", "banana", "broccoli", "raspberry")))

df %>% 
  mutate(category = fct_collapse(item,
                                 fruit = c("apple", "grape", "banana", "raspberry"),
                                 vegetables = "broccoli", group_other = TRUE))
#> # A tibble: 5 x 2
#>   item      category  
#>   <fct>     <fct>     
#> 1 apple     fruit     
#> 2 grape     fruit     
#> 3 banana    fruit     
#> 4 broccoli  fruit     
#> 5 raspberry vegetables
```
## Example 3 - with `group_other` argument set to `FALSE`
Just to demonstrate that this bug does not occur when `group_other` is set to `FALSE`:
```r
df <-
  tibble(item = factor(c("apple", "grape", "banana", "broccoli", "raspberry")))

df %>% 
  mutate(category = fct_collapse(item,
                                 fruit = c("apple", "grape", "banana", "raspberry"),
                                 vegetables = "broccoli", group_other = FALSE))
#> # A tibble: 5 x 2
#>   item      category  
#>   <fct>     <fct>     
#> 1 apple     fruit     
#> 2 grape     fruit     
#> 3 banana    fruit     
#> 4 broccoli  vegetables
#> 5 raspberry fruit

```
## Session Info
```r
sessionInfo()
#> R version 3.5.2 (2018-12-20)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 10 x64 (build 17134)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United Kingdom.1252 
#> [2] LC_CTYPE=English_United Kingdom.1252   
#> [3] LC_MONETARY=English_United Kingdom.1252
#> [4] LC_NUMERIC=C                           
#> [5] LC_TIME=English_United Kingdom.1252    
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] bindrcpp_0.2.2     tibble_2.0.1       dplyr_0.7.8       
#> [4] forcats_0.4.0.9000
#> 
#> loaded via a namespace (and not attached):
#>  [1] Rcpp_1.0.0       knitr_1.21       bindr_0.1.1      magrittr_1.5    
#>  [5] tidyselect_0.2.5 R6_2.3.0         rlang_0.3.1      fansi_0.4.0     
#>  [9] stringr_1.4.0    highr_0.7        tools_3.5.2      xfun_0.4        
#> [13] utf8_1.1.4       cli_1.0.1        htmltools_0.3.6  yaml_2.2.0      
#> [17] digest_0.6.18    assertthat_0.2.0 crayon_1.3.4     purrr_0.3.0     
#> [21] glue_1.3.0       evaluate_0.12    rmarkdown_1.11   stringi_1.2.4   
#> [25] compiler_3.5.2   pillar_1.3.1     pkgconfig_2.0.2
```

<sup>Created on 2019-02-17 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>
</details>

gtm19:
Another fix for #172, but where this time:

- `levs` is a list whether `group_other == TRUE` or not
- when the output does have an `"Other"` level, this level comes last (for consistency with `fct_other`)
- added tests to check for this functionality

```
library(forcats)

f1 <- factor(letters[1:4])
f2 <- fct_collapse(f1, x1 = c("a", "d"), x2 = "c", group_other = TRUE)

f2
[1] x1    Other x2    x1   
Levels: x1 x2 Other
```

Longer term I'm thinking it would be preferable for users to be able to specify the `Other` level, as is possible with `fct_other`.


batpigandme:
Also fixes #202

R/collapse.R
```diff
@@ -23,10 +23,14 @@ fct_collapse <- function(.f, ..., group_other = FALSE) {
     f <- check_factor(.f)
     levels <- levels(f)
     new[["Other"]] <- levels[!levels %in% levs]
-    levs <- levels
+    levs <- as.list(unlist(new, use.names = FALSE))
```
hadley: Can you explain what this is doing? Unlisting and relisting an object seems a bit weird.

R/collapse.R
```diff
@@ -23,10 +23,14 @@ fct_collapse <- function(.f, ..., group_other = FALSE) {
...
   }
 
   names(levs) <- names(new)[rep(seq_along(new), vapply(new, length, integer(1)))]
 
-  fct_recode(.f, !!!levs)
+  out <- fct_recode(.f, !!!levs)
+
+  if (any(levels(out) == "Other")){
+    fct_relevel(out, "Other", after = Inf)
```
hadley: Could we do this move earlier? (i.e. by adjusting `levs`?)

R/collapse.R
gtm19: It was just to flatten `new`, and I actually copied it from line 21 in the code (so you'll have to [ask yourself](https://github.com/tidyverse/forcats/blob/4866ef4295d99ed67024ac32ac6d742a407978df/R/collapse.R) :joy:). I did this for consistency but perhaps I should have thought about whether it was the best way.

As it happens, I believe there is no need for `new` to be `as.list`ed, and can just be flattened to a vector. Then, on line 26, `levs` can just be combined with itself and the strings which are going to be collapsed to `other` like so:

```suggestion
    levs <- c(levs, new[["Other"]])
```
Happy to make this change via pull request if it sounds reasonable.

R/collapse.R
gtm19: Without doing a deeper dive, I don't _think_ so.  The issue is that `check_factor` applies the default factor order of `base::factor`, meaning it will be a lottery as to whether 'Other' will appear last.

If it's the `if` operator which is bothersome, perhaps I could just delete the entire _if/else_ and manually provide the levels in the right order:
```
    levels(out) <- names(new)
```
If there is an _Other_, then it will already be last in `names(new)`. If not, nothing will change.

R/collapse.R
hadley: Yes, that looks better to be

R/collapse.R
gtm19: In short, I don't think so. I have reworked a little though, so it is consistent with `fct_other()`

tests/testthat/test-collapse.R
```diff
@@ -29,6 +29,27 @@ test_that("can collapse un-named levels to Other", {
   expect_equal(f2, factor(c("xy", "xy", "Other"), levels = c("xy", "Other")))
 })
 
+test_that("collapses levels correctly when group_other is TRUE but no Other variables to group", {
```
hadley:
```suggestion
test_that("collapses levels correctly when group_other is TRUE but no other variables to group", {
```

tests/testthat/test-collapse.R
```diff
@@ -29,6 +29,27 @@ test_that("can collapse un-named levels to Other", {
...
+})
+
+test_that("collapses levels correctly when group_other is TRUE and some Other variables to group", {
+  f1 <- factor(letters[1:4])
+  f2 <- fct_collapse(f1, x1 = c("a", "d"), x2 = "c", group_other = TRUE)
+
+  expect_equal(f2, factor(c("x1", "Other", "x2", "x1"), levels = c("x1", "x2", "Other")))
+})
+
+test_that("collapses levels correctly when group_other is FALSE", {
```
hadley: I think this is already handled by the existing tests so you can remove.

R/collapse.R
hadley: Ok, in that case, can you please fix the style: https://style.tidyverse.org/syntax.html#indenting

hadley:
Thanks @gtm19!

PR #178, opened 2019-02-20
add function first2

jtr13:
Adds a helpful function, `first2`, to `fct_reorder2` (as a complement to `last2`), which allows sorting on the first rather than the last value of `.x`. This is helpful when `.x` is a factor, and has other use cases, such as placing the legend on the left side of a multiline plot. 

Example:

```
library(tidyverse)
MASS::painters %>%
  rownames_to_column("Name") %>%
  gather(key = "Skill", value = "Value", -Name, -School) %>%
  ggplot(aes(Value, fct_reorder2(Name, Skill, Value, .fun = first2, .desc = FALSE),
             color = Skill)) + geom_point() +
  ggtitle("MASS::painters dataset", 
          sub = "Name sorted by Value for Skill==Colour (first factor level of Skill)")
```

image


hadley:
Looks great. Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jtr13:
Great, just added a bullet to NEWS. (This is my first tidyverse _code_ contribution... very exciting! 😄)

hadley:
Thanks @jtr13!

PR #189, opened 2019-03-20
Update fct_relevel.Rd

AndrewKinsman:


batpigandme:
Hi @AndrewKinsman,

Thanks for the fix. We actually use [roxygen2](https://github.com/klutometis/roxygen) to write the documentation right next to the R code, which is then extracted to generate the `.Rd` files. So, this will get overwritten next time we generate the documentation for `.Rd` (and for the html files for the pkgdown site) as it is right now.

Would you mind making the change in the `.R` file and adding it to this pull request? 

https://github.com/tidyverse/forcats/blob/master/R/relevel.R#L30

Let me know if you have any trouble.

Thanks!

batpigandme:
Closed in favour of #190

PR #200, opened 2019-07-08
Add description to both fct_lump and fct_lump_min

### LINKED ISSUE #195: fct_lump_min() needs better documentation
What does it actually do? Probably needs a sentence in the `@description`.

VincentGuyader:
Fixes #195

### DIFF
--- a/R/lump.R
+++ b/R/lump.R
@@ -1,6 +1,10 @@
-#' Lump together least/most common factor levels into "other"
-#' @description `fct_lump()` lumps together least/most common factor levels into "other".
-#'  `fct_lump_min()` lumps together factor which doesn't appear at least `min` number of times.
+#' Lump together factor levels into "other"
+#'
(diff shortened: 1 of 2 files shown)


hadley:
(FYI I changed your title and body to reflect our standard naming practices)

R/lump.R
```diff
@@ -1,5 +1,5 @@
 #' Lump together least/most common factor levels into "other"
-#'
+#' @description `fct_lump` lumps together least/most common factor levels into "other".
```
hadley: I think it's a bit easier to read if you just document both functions here — that way you can write a single paragraph. (It is a bit more confusing when you look at the docs for `fct_lump_min()` but I think that trade off is worth it).

Also can you please use parens at the end of function names? (https://style.tidyverse.org/documentation.html#r-code)

VincentGuyader: I have made changes that take into account these remarks

Regards

hadley:
Thanks!

PR #201, opened 2019-07-08
as_factor() example for numeric-method object

### LINKED ISSUE #198: Add numeric-method example to `as_factor()`
Now that `as_factor()` has a numeric method (thanks to https://github.com/tidyverse/forcats/pull/170), it might be worth adding a simple example of this (e.g. the example used in the forcats 0.4.0 release post https://github.com/tidyverse/tidyverse.org/blob/a117fc4fb543f89282a0f8edb38a8fd5d59b8b90/content/articles/2019-05-forcats-0-4-0.Rmarkdown#L73-L78 )

sinarueeger:
1. Added a third example to `as_factor()` that shows the behaviour with a numeric vector. 

```
y <- c("1.1", "11", "2.2", "22")
z <- as_numeric(y)
as_factor(z)
as.factor(z)
```

Example copied from https://github.com/tidyverse/tidyverse.org/blob/a117fc4fb543f89282a0f8edb38a8fd5d59b8b90/content/articles/2019-05-forcats-0-4-0.Rmarkdown#L73-L78. 

2. Added short descriptions to each example.

~3. Updated `NEWS.md`.~

4. Limited current function description _"creates levels in the order in which they appear"_ to characters and clarified behaviour with numerics (#170) (_"When x is numeric, the ordering is based on the numeric value and consistent with base R."_).

Closes #198.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 * `first2()`, a `fct_reorder2()` helper function, sorts `.y` by the first value of `.x` (@jtr13).
 
+## Minor bug fixes and improvements
+
+* Added numeric-method example to `as_factor()` (@sinarueeger, #198).
+
 # forcats 0.4.0
 
 ## New features
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,10 +2,6 @@
 
 * `first2()`, a `fct_reorder2()` helper function, sorts `.y` by the first value of `.x` (@jtr13).
 
-## Minor bug fixes and improvements
-
-* Added numeric-method example to `as_factor()` (@sinarueeger, #198).
-
 # forcats 0.4.0
 
 ## New features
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 * `first2()`, a `fct_reorder2()` helper function, sorts `.y` by the first value of `.x` (@jtr13).
 
+## Minor bug fixes and improvements
```
hadley: Since this is just a documentation improvement (and not a user facing change), it doesn't need a news bullet.

hadley: Otherwise, looks great!

DavisVaughan:
https://github.com/tidyverse/forcats/pull/170 explicitly states "By default, orders factors in numeric order, unlike the other methods which default to order of appearance."

This is different from what the description in the documentation states: "creates levels in the order in which they appear".

While you are documenting the numeric method, it might also be worth updating the description of the function to say something about this difference.

NEWS.md
sinarueeger: I removed the news bullet.

sinarueeger:
@DavisVaughan Updated the PR description.

Issue #194, opened 2019-04-02
fct_collapse() - change "Other" name


jamiefo:
Hi,

For the fct_collapse() function there is an option to set group_other = TRUE/FALSE.
May there please be an option added to rename the "Other" group to anything we want so that it will minic a case_when statement.  It's inefficient to use both fct_collapse() and then fct_other() to rename the "Other" bucket.  I would want to name the catch all bucket to something of my choice.  This is a bit of a pain point for me when using it.  I believe most would run to a case_when() statement if this option was not updated.

james

gtm19:
I did consider incorporating this into my fix (#176) for #172, but couldn't decide whether this flexibility should be confined to `fct_other()`.

@hadley?

hadley:
@gtm19 I think it would be best as a separate PR. Maybe we could change the default of `group_other` to `NULL`, and then when it's a string we'd use that as the name of the other group? (We'd also need to check for `TRUE` to make sure it's backward compatible)

gtm19:
@hadley Either that, or we could add an `other_level` argument which defaults to `"Other"` (for consistency with `fct_other()`). 

I think I prefer the latter, and this would still be backwards compatible.

hadley:
Two arguments seem excessive for this simple feature. Maybe we should just use `other_level` and deprecate `group_other`?

PR #205, opened 2019-08-20
fix fct_collapse group_other = TRUE misleveling

avishaitsur:
This fix addresses the wrong leveling that occurred due to assigning the old values to the new factor instead of both the new levels and the unchanged ones (who then become "Other"). As far as I've tried it, it works... but I'm no R expert and also new to GitHub as well - just uses fct_collapse quite a lot.

### DIFF
--- a/R/collapse.R
+++ b/R/collapse.R
@@ -18,12 +18,12 @@
 #' fct_count(partyid2)
 fct_collapse <- function(.f, ..., group_other = FALSE) {
   new <- rlang::dots_list(...)
-  levs <- as.list(unlist(new, use.names = FALSE))
+  levs <- unlist(new, use.names = FALSE)
   if (group_other){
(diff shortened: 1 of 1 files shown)


batpigandme:
This is actually already covered in #176, but hasn't been merged yet.

avishaitsur:
Oh, I see -
I've seen the fix by gtm19 is from Feb 18th, 2019,
but the conversation thread from May 10th, 2019 said to drop fct_collapse so I thought there is some kind of a problem and still prevents it from working.

Would it be merged soon?

batpigandme:
> Would it be merged soon?

I imagine so.

avishaitsur:
Thanks for the reply.

Also added a name_other argument to allow to name the "Other" category differently and address issue #194 
Defaults to "Other" so should have backward compatibility problems.

Issue #209, opened 2019-09-23
Feature request: .desc  = FALSE option for fct_infreq()


jtr13:
For horizontal bar charts, it is better to sort by increasing order of frequency, since bars are plotted bottom to top. A `.desc = FALSE` option in `fct_infreq()` would be cleaner than using `fct_rev()`:

df <- tibble(x = c("red", "red", "red", "yellow", "yellow", "blue"))
ggplot(df, aes(fct_rev(fct_infreq(x)))) + geom_bar() + coord_flip()

![image](https://user-images.githubusercontent.com/14161912/65453831-f5bb1680-de11-11e9-93cd-6a04d2e0cc2d.png)

jtr13:
This has been addressed in pull request #210

hadley:
I don't think an argument is the right solution because it would result in more net typing:

```R
ggplot(df, aes(fct_rev(fct_infreq(x)))) + geom_bar() + coord_flip()
ggplot(df, aes(fct_infreq(x, .desc = TRUE))) + geom_bar() + coord_flip()
```

PR #210, opened 2019-09-24
add .desc parameter to fct_infreq with default TRUE

jtr13:
It is helpful for horizontal bar charts to be able to set `.desc = FALSE` in `fct_infreq()`: see #209

### DIFF
--- a/R/reorder.R
+++ b/R/reorder.R
@@ -108,10 +108,10 @@ fct_inorder <- function(f, ordered = NA) {
 
 #' @export
 #' @rdname fct_inorder
-fct_infreq <- function(f, ordered = NA) {
+fct_infreq <- function(f, ordered = NA, .desc = TRUE) {
   f <- check_factor(f)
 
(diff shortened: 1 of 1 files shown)


R/reorder.R
```diff
@@ -108,10 +108,10 @@ fct_inorder <- function(f, ordered = NA) {
 
 #' @export
 #' @rdname fct_inorder
-fct_infreq <- function(f, ordered = NA) {
+fct_infreq <- function(f, ordered = NA, .desc = TRUE) {
```
hadley: This should be called `desc` I think, and can you please document it?

hadley: Hmmm although then you'd call it `fct_infreq(x, desc = FALSE)` and the meaning doesn't seem that clear to me?

hadley:
Hmmm, let me contemplate this issue more. I'm not sure that using `fct_rev()` isn't the best solution.

Issue #185, opened 2019-03-01
Default factor levels when coercing from logical


alejandroschuler:
Hi folks, I'm not sure if `forcats` is the right place to raise this issue, so apologies in advance if this should go elsewhere.

I find it curious that `factor(c(TRUE, TRUE))` produces a factor with a single level: `TRUE`. If we are to imagine that factors are categories where we know all of the levels a-priori, wouldn't it make sense if logicals converted to factors preserved that a-priori knowledge of the two levels `TRUE` and `FALSE` (with a consistent order) regardless of whether or not one of those two levels does not appear in the data? It's a trivial fix using `levels=c("TRUE", "FALSE")`, but I find it to be so philosophically obvious that this should always happen that it should warrant default behavior. Perhaps something along the lines of:

```r
fct.default = base::factor
fct.logical = function(lgl, ...) base::factor(lgl, levels=c("TRUE", "FALSE"))
fct = function(x, ...) UseMethod("fct")
```

This is a sometimes a minor frustration for me when using `tidymodels` packages (and `caret` before that) which require a factor outcome for classification problems. Every now and again I end up in a corner case where I get errors like:

```r
 Error: In metric: `yardstick::sens`
`truth` and `estimate` levels must be equivalent.
`truth`: TRUE, FALSE
`estimate`: TRUE
```

hadley:
Makes sense, but what about `NA`? There are three possible values of a logical vector after all.

alejandroschuler:
NA behavior is the same as with `base::factor`.
```r
> fct(c(TRUE, FALSE, NA))
[1] TRUE  FALSE <NA> 
Levels: TRUE FALSE
> base::factor(c(TRUE, FALSE, NA))
[1] TRUE  FALSE <NA> 
Levels: FALSE TRUE
```
As is the case with other factors, `NA`s are represented as `NA`s, not as an explicit level. I would say this is the desired default behavior because although `NA` is a possible value that the variable can take _in the data_, it is not a value the variable can take _epistemologically_. If that behavior is desired, `fct_explicit_na` does the job as usual.

alejandroschuler:
Perhaps the best place to implement this in `forcats` would be as a method of `as_factor`:
```r
as_factor.logical = function(lgl, ...) base::factor(lgl, levels=c("TRUE", "FALSE"), ...)
```

alejandroschuler:
[...]

PR #212, opened 2019-09-26
add logical method to as_factor

### LINKED ISSUE #211: add a logical method to as_factor
The goal is to avoid this kind of error message :

``` r
forcats::as_factor(logical())
#> Error in UseMethod("as_factor"): pas de méthode pour 'as_factor' applicable pour un objet de classe "logical"
```

<sup>Created on 2019-09-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I think that a good way to solve this is to use : 

```
as_factor.logical <- function(x,...){
  as_factor(as.character(x))
}
```

VincentGuyader:
PR to fix #211 

(sorry some commit came from #195 ... )

### DIFF
--- a/R/as_factor.R
+++ b/R/as_factor.R
@@ -42,3 +42,12 @@ as_factor.character <- function(x, ...) {
 as_factor.numeric <- function(x, ...) {
   factor(x)
 }
+
+
+
+
+#' @rdname as_factor
+#' @export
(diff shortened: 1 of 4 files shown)


hadley:
Would you mind implementing along the lines of #185?

PR #220, opened 2019-11-18
Add fct_reordern() (Fix #16)

### LINKED ISSUE #16: Feature request: Order factor by (multiple functions of) multiple variables
For plotting and tables, it’s useful to reorder levels of a factor according to other variables, first by one variable, and then by other variables to break any ties. The summary function used for each variable may be different. Example:

```r
d = data.frame(
  name = factor(c("A", "A", "B", "B", "C", "C", "C", "D"),
                levels=LETTERS[4:1]),
  quality = c(5, 3, 4, 4, 7, 7, 7, 7),
  year = c(2000, 2001, 2013, 2014, 2015, 2015, 2015, 2015),
  weight = c(50, 45, 60, 57, 47, 50, 500, 63)
)
```

I want to order the `name` factor by 1) average quality, then 2) first `year` the product appeared, and then 3) median weight. If any ties remain, keep the original label order for these ties. In this example, the levels would be ordered ABCD.

To do this reordering, I have to think backwards, reordering by the last tie-breaker first, using either `fct_reorder()` or `reorder()`:

```r
d$name = fct_reorder(d$name, d$weight, median)
d$name = fct_reorder(d$name, d$year, min)
d$name = fct_reorder(d$name, d$quality, mean)
```

I would be very convenient to be able to do this in one go, using something like this (I’m dropping the d$ prefix to make the code clearer):

```r
name = fct_reordern(name,
                    vars = list(quality, year, weight),
                    funs = list(mean, min, median))
```

It would be even better if the functions were shown along with the variable names, e.g.:

```r
name = fcr_reordern(name,
                    mean(quality), min(year), median(weight))
```

For descending order, perhaps by using a `desc()` function, like this?

```r
name = fcr_reordern(name,
                    desc(mean(quality)), min(year), median(weight))
```

billdenney:
Fix #16 

@hadley, I saw what you meant in the #16 discussion about affecting the value and not just the levels.  That is addressed here.  Also, I removed the `.desc` argument.  Since it was both simpler and potentially friendlier to the user to leave the `decreasing` argument from `base::order()` exposed (it doesn't have the `dplyr` dependency to sort decreasing), I left that.  And, I made the default `method` "radix" so that it would work with a vector of values for `decreasing` (something that would annoy me whenever I forgot that I needed to set the method, and it seemed like a minor wart to add an argument directly).

Anyway, let me know what you think and if I'm missing something else important.


NEWS.md
```diff
@@ -1,46 +1,52 @@
...
+
+* `fct_reordern()` is a new function to order based on an arbitrary number of
+  values (@billdenney, #16)
 
 # forcats 0.4.0
 
 ## New features
 
-* `fct_collapse()` gains a `group_other` argument to allow you to group all 
+* `fct_collapse()` gains a `group_other` argument to allow you to group all
```
hadley: You seem to have changed multiple lines in this file. Can you please revert?

R/reorder.R
```diff
@@ -68,6 +69,30 @@ fct_reorder2 <- function(.f, .x, .y, .fun = last2, ..., .desc = TRUE) {
...
+#' 
+#' @param .f A factor (or character vector).
+#' @param method Passed to \code{base::order()}
+#' @param ordered Passed to \code{\link{fct_inorder}()}
+#' @inheritDotParams base::order -method
+#' @family Reordering
+#' @examples
+#' A <- c(3, 3, 2, 1)
+#' B <- c("A", "B", "C", "D")
+#' fct_reordern(.f = c("A", "B", "C", "D"), A, B)
```
hadley: Please don't name the first argument in calls.

R/reorder.R
```diff
@@ -68,6 +69,30 @@ fct_reorder2 <- function(.f, .x, .y, .fun = last2, ..., .desc = TRUE) {
...
+#' fct_reordern(.f = c("A", "B", "C", "D"), A, B, decreasing = TRUE)
```
hadley: Per my comment in the issue, I would suggest not providing this argument and instead assuming that the user will reverse ordering in some other way (e.g. using `dplyr::desc()`)

R/reorder.R
```diff
@@ -68,6 +69,30 @@ fct_reorder2 <- function(.f, .x, .y, .fun = last2, ..., .desc = TRUE) {
...
+#' fct_reordern(.f = c("A", "B", "C", "D"), A, B, decreasing = c(FALSE, TRUE))
+#' @export
+fct_reordern <- function(.f, ..., ordered = NA, method="radix") {
+  f <- check_factor(.f)
+  new_order <- base::order(..., method=method)
+  f_sorted <- f[new_order]
+  # Don't keep the value order for new_f as they will be inaccurately sorted.
+  # Just keep the levels and ordered state.
+  new_f <- fct_inorder(f_sorted, ordered = ordered)
```
hadley: I don't understand the logic here. Can you please use one of the existing reordering functions (e.g. `forcats::lvls_reorder()`)?

R/reorder.R
```diff
@@ -68,6 +69,30 @@ fct_reorder2 <- function(.f, .x, .y, .fun = last2, ..., .desc = TRUE) {
...
+#' @examples
+#' A <- c(3, 3, 2, 1)
+#' B <- c("A", "B", "C", "D")
+#' fct_reordern(.f = c("A", "B", "C", "D"), A, B)
+#' fct_reordern(.f = c("A", "B", "C", "D"), A, B, decreasing = TRUE)
+#' fct_reordern(.f = c("A", "B", "C", "D"), A, B, decreasing = c(FALSE, TRUE))
+#' @export
+fct_reordern <- function(.f, ..., ordered = NA, method="radix") {
+  f <- check_factor(.f)
+  new_order <- base::order(..., method=method)
```
hadley: I'd suggest not supply `method` to avoid the R 3.2 issue.

tests/testthat/test-reorder.R
```diff
@@ -57,3 +57,52 @@ test_that("fct_inseq gives error for non-numeric levels", {
...
+  )
+  expect_equal(
+    fct_reordern(.f = f, A, B, decreasing = c(FALSE, TRUE)),
+    factor(f, levels = c("D", "C", "B", "A"))
+  )
+  expect_equal(
+    fct_reordern(.f = f, A, B, decreasing = c(FALSE, TRUE)),
+    factor(f, levels = c("D", "C", "B", "A"))
+  )
+  # Checks of ordering
```
hadley: Given that the ordering is handled by an existing function, you only need a single test to make sure all the arguments are passed through appropriately.

tests/testthat/test-reorder.R
```diff
@@ -57,3 +57,52 @@ test_that("fct_inseq gives error for non-numeric levels", {
...
+test_that("fct_reordern works for all scenarios (fix #16)", {
+  A <- c(3, 3, 2, 1)
+  B <- c("A", "B", "C", "D")
+  f <- c("A", "B", "C", "D")
+  f_factor_ordered <- factor(f, ordered = TRUE)
+  expect_equal(
+    fct_reordern(.f = f, A, B),
+    factor(f, levels = c("D", "C", "A", "B"))
+  )
+  expect_equal(
```
hadley: You can remove the tests for `decreasing = TRUE` because you are just testing the behaviour of an existing function.

NEWS.md
billdenney: Sure.  As an FYI, the changes were made by my default emacs settings which remove blank space at the end of lines, and then I tried to make the style match the tidyverse style guide on line wrapping (https://style.tidyverse.org/news.html#bullets).

R/reorder.R
billdenney: Fixed for all examples.

R/reorder.R
billdenney: Removed.

R/reorder.R
billdenney: Ah, I didn't know of `lvls_reorder()`.  Simplified.

R/reorder.R
billdenney: Removed.

tests/testthat/test-reorder.R
billdenney: Removed except for one to confirm that interaction between the two works as decreasing sorting is an important behavior.

hadley: That adds a dependency on dplyr — instead you'll need to implement a local version of `desc()` inside the test, e.g. `desc <- function(x) -xtfrm(x)`

NEWS.md
hadley: As a general rule, when working with someone else's code, you should suppress all the automated formatting that you might usually do.

tests/testthat/test-reorder.R
billdenney: For my education, I thought that since the dplyr call is in a test and dplyr is already in the DESCRIPTION Suggests field, I thought that this was OK.  (Similarly, since I have it in the examples in the help, I thought that the same applied there.)  I'm happy to implement that way, but my hope was that this would be simpler for the user (via help) and future maintainer (via tests) to use.

Whichever way you confirm, I'll update to that.

NEWS.md
billdenney: Fair enough.

tests/testthat/test-reorder.R
billdenney: Updated.

billdenney:
@hadley, I think that I have addressed everything.

billdenney:
FYI, I have re-updated to current main and fixed all CI issues (mainly the updated issue around pkgdown changes since this PR was originally proposed).

Issue #182, opened 2019-02-26
fct_cross() fails if its inputs are 3 or more factors


ghost:
```r
library(forcats)
fruit <- factor(c("apple", "kiwi", "apple", "apple"))
colour <- factor(c("green", "green", "red", "green"))
eaten <- factor(c("yes", "no", "yes", "no"))

fct_cross(fruit, colour, eaten)
#> Error: Column name `<fct>` must not be duplicated.
#> Use .name_repair to specify repair.
```

This issue also happens if there is more than 1 character vector included as input.

It seems to me that the issue gets fixed if the following line: 
```.data <- lapply(tibble::tibble(.f, !!!flist), check_factor)```

is changed to this:
```.data <- lapply(rlang::list2(.f, !!!flist), check_factor)```

ibwoo:
This issue helped me immensely - I hadn't cottoned on to the fact that they couldn't all be factors. An easy enough workaround for the moment, but this does deserve to be fixed!

PR #224, opened 2019-12-20
Fix multiple factors issue

kbodwin:
`fct_cross` was breaking with more than one additional factor, because the default variable name for `tibble::tibble()` is "<fct>", and so there were duplicate variable names.  I've added the option `.name_repair` to take care of this, and suppressed the resulting message.

Addresses #182

### DIFF
--- a/R/fct_cross.R
+++ b/R/fct_cross.R
@@ -2,8 +2,7 @@
 #'
 #' Computes a factor whose levels are all the combinations of the levels of the input factors.
 #'
-#' @param .f A factor or character vector
-#' @param ...  Additional factors or character vectors
+#' @param ...  Any number of factors or character vectors
 #' @param sep A character string to separate the levels
 #' @param keep_empty If TRUE, keep combinations with no observations as levels
 #' @return The new factor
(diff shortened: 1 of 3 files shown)


R/fct_cross.R
```diff
@@ -24,7 +24,8 @@ fct_cross <- function(.f, ..., sep = ":", keep_empty = FALSE) {
     return(.f)
   }
 
-  .data <- lapply(tibble::tibble(.f, !!!flist), check_factor)
+  .data <- suppressMessages(tibble::tibble(.f, !!!flist, .name_repair = "universal"))
```
hadley: Hmmmmm, why does it get put into a tibble in the first place? I don't think the names are ever actually used?

hadley:
(BTW you need to use specifically `Fixes #182` or `Closes #182` to activate the github magic that will cause the issue to be closed when the PR is merged)

R/fct_cross.R
kbodwin: I'm not totally sure, but I think it's just a convenience for getting all the factors together when the length of `flist` is unknown.

I actually think it makes the most sense to just not have `.f` as a separate input...

kbodwin:
> (BTW you need to use specifically `Fixes #182` or `Closes #182` to activate the github magic that will cause the issue to be closed when the PR is merged)

Thank you!  This should be in the PR description or in the commit message or both?

Regardless, I believe the above commit Fixes #182

hadley:
Would you mind also providing a unit test that checks that crossing 2+ factors now works?

R/fct_cross.R
hadley: It was actually needed for recycling, so I've bought it back with a test.

kbodwin:
Incorporated changes into #236

Issue #225, opened 2019-12-20
fct_cross does not preserve level ordering


kbodwin:
Very related to #183 .

`fct_cross` relevels factors after pasting, thus ordering them in alphabetical order.  

When you set `keep_empty = TRUE`, however, the original levels are preserved.  

``` r
library(forcats)

f1 <- c("a", "b", "c")
f2 <- c("alpha", "beta", "gamma")

f1 <- factor(f1, levels = c("b", "c", "a"))

fct_cross(f1, f2)
#> [1] a:alpha b:beta  c:gamma
#> Levels: a:alpha b:beta c:gamma

fct_cross(f1, f2, keep_empty = TRUE)
#> [1] a:alpha b:beta  c:gamma
#> 9 Levels: b:alpha c:alpha a:alpha b:beta c:beta a:beta ... a:gamma
```

<sup>Created on 2019-12-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

hadley:
I think this is just a duplicate of #183

Issue #183, opened 2019-02-26
fct_cross() provides different order of levels depending on keep_empty logical


ghost:
Sorry, I don't mean to pick on that function along with #182.

`fct_cross()` resets the order of the levels of the original factor variables when the `keep_empty` argument is kept to FALSE, but maintains the original order if the argument is set to TRUE.

``` r
library(forcats)

fruit <- factor(c("apple", "kiwi", "apple", "apple"), levels = c("kiwi", "apple"))
colour <- factor(c("green", "green", "red", "green"), levels = c("red", "green"))

new_f <- fct_cross(fruit, colour, keep_empty = FALSE)
levels(new_f)
#> [1] "apple:green" "apple:red"   "kiwi:green"

new_f2 <- fct_cross(fruit, colour, keep_empty = TRUE)
levels(new_f2)
#> [1] "kiwi:red"    "apple:red"   "kiwi:green"  "apple:green"
```

I've re-worked the function below so that the order of the original factor variables is maintained regardless of the keep_empty argument (note that I've included the suggested fix in #182).

```r
new_fct_cross <- function(.f, ..., sep = ":", keep_empty = FALSE) {
  .f <- check_factor(.f)
  
  flist <- rlang::list2(...)
  if (length(flist) == 0) {
    return(.f)
  }
  
  .data <- lapply(rlang::list2(.f, !!!flist), check_factor)
  newf <- rlang::invoke(paste, .data, sep = sep)
  
  all_levels <- lapply(.data, levels)
  return_f <- factor(newf, levels = rlang::invoke(paste,
                                      rlang::invoke(expand.grid, all_levels),
                                      sep = sep
  ))
  
  if (keep_empty) {
    return_f
  } else {
    fct_drop(return_f)
  }
}

fruit <- factor(c("apple", "kiwi", "apple", "apple"), levels = c("kiwi", "apple"))
colour <- factor(c("green", "green", "red", "green"), levels = c("red", "green"))

new_f <- new_fct_cross(fruit, colour, keep_empty = FALSE)
levels(new_f)
#> [1] "apple:red"   "kiwi:green"  "apple:green"

new_f2 <- new_fct_cross(fruit, colour, keep_empty = TRUE)
levels(new_f2)
#> [1] "kiwi:red"    "apple:red"   "kiwi:green"  "apple:green"
```

PR #226, opened 2019-12-20
Fct_cross level ordering

kbodwin:
Addresses #183 and #225 

Added the option `preserve_orders` to  `fct_cross`, for when you want the cross-product levels to respect the original level ordering instead of refactoring to alphabetical order.

For example (and also a common use case):

``` r
library(forcats)

f1 <- c("Fall", "Fall", "Winter", "Spring")
f2 <- c("2019", "2018", "2018", "2018")

f1 <- factor(f1, levels = c("Fall", "Winter", "Spring"))

fct_cross(f1, f2)
#> [1] Fall:2019   Fall:2018   Winter:2018 Spring:2018
#> Levels: Fall:2018 Fall:2019 Spring:2018 Winter:2018

fct_cross(f1, f2, preserve_orders = TRUE)
#> [1] Fall:2019   Fall:2018   Winter:2018 Spring:2018
#> Levels: Fall:2018 Winter:2018 Spring:2018 Fall:2019
```

<sup>Created on 2019-12-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

It's my opinion that the ordering should be respected by default.  However, I've left the default to as `preserve_orders = FALSE` for the time being, for backwards compatibility reasons.

### DIFF
--- a/R/fct_cross.R
+++ b/R/fct_cross.R
@@ -16,7 +16,7 @@
 #' fct_cross(fruit, colour)
 #' fct_cross(fruit, colour, eaten)
 #' fct_cross(fruit, colour, keep_empty = TRUE)
-fct_cross <- function(.f, ..., sep = ":", keep_empty = FALSE) {
+fct_cross <- function(.f, ..., sep = ":", keep_empty = FALSE, preserve_orders = FALSE) {
   .f <- check_factor(.f)
 
   flist <- rlang::list2(...)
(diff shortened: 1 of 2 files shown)


R/fct_cross.R
```diff
@@ -16,7 +16,7 @@
 #' fct_cross(fruit, colour)
 #' fct_cross(fruit, colour, eaten)
 #' fct_cross(fruit, colour, keep_empty = TRUE)
-fct_cross <- function(.f, ..., sep = ":", keep_empty = FALSE) {
+fct_cross <- function(.f, ..., sep = ":", keep_empty = FALSE, preserve_orders = FALSE) {
```
hadley: I think this should just be the default, and it's not clear to me that it even needs to be an argument?

R/fct_cross.R
```diff
@@ -33,6 +33,13 @@ fct_cross <- function(.f, ..., sep = ":", keep_empty = FALSE) {
       rlang::invoke(expand.grid, all_levels),
       sep = sep
     ))
+  } else if (preserve_orders) {
+    all_levels <- lapply(.data, levels)
```
hadley: I think the way to fix this is to generate the full set of levels outside the if the statement, and the filter them to the set of actual levels (using `intersect()`, which is a shortcut for your `%in%` pattern) if `keep_empty` is `FALSE`.

kbodwin:
incorporated changes into #236

PR #237, opened 2020-01-31
Change `as_factor()` to convert ordered factors to plain factors

### LINKED ISSUE #216: Feature Request: handle ordered factors within as_factor()
Currently, there is not a way to convert an `"ordered" "factor"` into a plain `"factor"` within `{forcats}`.

Hadley requested that PR https://github.com/tidyverse/forcats/pull/215#issuecomment-542714204 be converted into an issue. For easy reading, here's the comment:

> Based on our current understanding of S3, I think it's clear that these functions should be as.factor() and as.ordered(). as.ordered() works currently, but as.factor() does not. We can't fix as.factor(), but we can fix as_factor(). Doing that 100% correctly is a bit fiddly, so would you mind filing an issue?

robinsones:
Closes #216. 

I used `structure()` as that's what `as_factor.character` uses. Added two tests and a NEWS item as well.

### DIFF
--- a/R/as_factor.R
+++ b/R/as_factor.R
@@ -36,8 +36,8 @@ as_factor <- function(x, ...) {
 as_factor.factor <- function(x, ...) {
   structure(
     x,
-    label = attr(x, "label", exact = TRUE),
     class = "factor",
+    label = attr(x, "label", exact = TRUE),
     levels = attr(x, "levels", exact = TRUE)
   )
 }
--- a/R/as_factor.R
+++ b/R/as_factor.R
@@ -34,7 +34,12 @@ as_factor <- function(x, ...) {
 #' @rdname as_factor
 #' @export
 as_factor.factor <- function(x, ...) {
-  x
+  structure(
+    x,
+    label = attr(x, "label", exact = TRUE),
+    class = "factor",
+    levels = attr(x, "levels", exact = TRUE)
+  )
 }
 
 #' @rdname as_factor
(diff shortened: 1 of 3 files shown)


R/as_factor.R
```diff
@@ -34,7 +34,12 @@ as_factor <- function(x, ...) {
 #' @rdname as_factor
 #' @export
 as_factor.factor <- function(x, ...) {
-  x
+  structure(
+    x,
+    label = attr(x, "label", exact = TRUE),
```
hadley: Can you please order these alphabetically?

tests/testthat/test-as_factor.R
```diff
@@ -16,3 +16,10 @@ test_that("supports NA (#89)", {
   x <- c("a", "z", "g", NA)
   expect_equal(as_factor(x), fct_inorder(x))
 })
+
+test_that("converts ordered factor into plain factor (#216)", {
+  x <- factor(c("a", "z", "g"), ordered = TRUE)
+  x <- as_factor(x)
+  expect_false(is.ordered(x))
+  expect_equal(attributes(x)$class, "factor")
```
hadley: I think I'd just do `expect_s3_class(as_factor(x), "factor", exact = "TRUE")` (i.e. you only need one test)

R/as_factor.R
robinsones: done

tests/testthat/test-as_factor.R
robinsones: done

hadley:
Thanks @robinsones!

hadley:
Unfortunately I had to revert this PR because it causes haven to fail R CMD check. That's totally a bug with haven, but because haven has C warnings that are slightly outside of my control, it's unfortunately easier to revert here than to squeeze out a quick haven release.

Issue #233, opened 2020-01-21
Import rlang


hadley:
Add `#' @import rlang` to `forcats-package.R` and then remove all uses of `rlang::`

(This makes forcats more consistent with how we use rlang in other packages)

PR #240, opened 2020-01-31
Rlang call update

gralgomez:
Worked on issue #233 with @gabrielodom


.Rbuildignore
```diff
@@ -16,3 +16,6 @@
 ^revdep$
 ^\.github$
 ^LICENSE\.md$
+^forcats\.Rcheck$
```
hadley: It's not a big deal to add these, but it suggests something might be wonky with your process, as they should end up in the parent directory, not here.

NEWS.md
```diff
@@ -6,6 +6,8 @@
   
 * fixed bug in `fct_collapse()` so it now correctly collapses factors when `group_other = TRUE` (#172), and makes `"Other"` the last level (#202) (@gtm19, #172 & #202)
 
+* Updated calls to **rlang** package (#233). Package called globally with `@import rlang` instead of `rlang::` calls. Removed all `rlang::` calls (@gralgomez, @gabrielodom, #233)
```
hadley: For future reference, you don't need a news bullet here since it doesn't affect the user of the package.

.Rbuildignore
gralgomez: Ok! I'm running it from RStudio Cloud. Thank you for the tip!

NEWS.md
gralgomez: Gotcha! Thanks

hadley:
Thank you @gralgomez and @gabrielodom!

hadley:
I think you can move your issue directly to gonged 😄

PR #241, opened 2020-01-31
Split fct_lump components to fct_prop and fct_count

### LINKED ISSUE #142: fct_lump all with count above/below threshold
It would be great if `fct_lump()` had an additional mode that lumped factor levels not based on the final number of factor levels desired (`n`) or the _proportion_ of the data that accrues to the level (`prop`) but based on the _raw count_ for the level (call it `count`?).

Stealing the example data from https://github.com/tidyverse/forcats/issues/43:

    N <- 5
    M <- 2 ** N
    exp_factor <- factor(rep(1:(N + M), c(2 ** (N:1), rep(1, M))))
    table(exp_factor)
    ## exp_factor
    ##  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 
    ## 32 16  8  4  2  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1 
    ## 26 27 28 29 30 31 32 33 34 35 36 37 
    ##  1  1  1  1  1  1  1  1  1  1  1  1

Desired behavior: `forcats::fct_lump(exp_factor, count = 1)` would retain only the factor levels `1:5` and `Other`.

### LINKED ISSUE #167: Splitting up fct_lump()
As discussed in #142, worth considering if `fct_lump()` should be split up into `fct_lump_prop()` and `fct_lump_n()`. `fct_lump_min()` has been been created.

jonocarroll:
* Splits out some functionality to helper functions, which are then used for new `fct_lump_prop` and `fct_lump_n`.
* Refactors `fct_lump_min` to use new internal functionality.
* New tests, and all pass with one unrelated warning

```
test-fct_cross.R:6: warning: gives correct levels
`invoke()` is deprecated as of rlang 0.4.0.
Please use `exec()` or `eval(expr())`instead.
This warning is displayed once per session.
```

* `fct_lump` retains `prop` and `n` arguments but now these routes have their own dedicated functions with shared internals.

Closes #167 
Closes #142


R/lump.R
```diff
@@ -4,6 +4,8 @@
 #' `prop`) into "other".
 #' `fct_lump_min()` lumps together all levels which don't appear at least
 #' `min` number of times.
+#' `fct_lump_prop` lumps together all levels which appear at or below a proportion of `prop`.
```
hadley: Can you turn this into a bulleted list please?

hadley: And the function names should have `()` after them

R/lump.R
```diff
@@ -137,6 +185,54 @@ fct_lump_min <- function(f, min, w = NULL, other_level = "Other") {
 
 }
 
+.fct_lump_min <- function(min, count, levels, other_level) {
```
hadley: I think pulling these functions out make the code harder to understand because you've added an additional layer of indirection — I don't think it's worth it given that they're each only called from one place.

R/lump.R
```diff
@@ -56,46 +58,33 @@
 #'
 fct_lump <- function(f, n, prop, w = NULL, other_level = "Other",
                      ties.method = c("min", "average", "first", "last", "random", "max")) {
-  f <- check_factor(f)
-  w <- check_weights(w, length(f))
-  ties.method <- match.arg(ties.method)
 
+  ties.method <- match.arg(ties.method)
+  calcs <- calc_levels(f, w)
```
hadley: I think it would be clearer to keep the variables in `calcs` (i.e. don't unpack them)

R/lump.R
```diff
@@ -56,46 +58,33 @@
...
   if (!xor(missing(n), missing(prop))) {
     new_levels <- ifelse(!in_smallest(count), levels, other_level)
   } else if (!missing(n)) {
-    if (n < 0) {
-      rank <- rank(count, ties = ties.method)
-      n <- -n
-    } else {
-      rank <- rank(-count, ties = ties.method)
-    }
+    check_rank <- .fct_lump_n(n, count, levels, other_level, ties.method)
```
hadley: Why doesn't this call `fct_lump_count()`?

R/lump.R
```diff
@@ -56,46 +58,33 @@
...
+    } else {
+      new_levels <- check_rank$new_levels
     }
 
-    new_levels <- ifelse(rank <= n, levels, other_level)
   } else if (!missing(prop)) {
-    prop_n <- count / total
-    if (prop < 0) {
-      new_levels <- ifelse(prop_n <= -prop, levels, other_level)
+    check_prop <- .fct_lump_prop(prop, count, total, levels, other_level)
```
hadley: Similarly, why doesn't this call `fct_lump_prop()`? I think `fct_lump()` should just become a wrapper that automatically picks which lower level function you want.

R/lump.R
jonocarroll: `.fct_lump_prop` and `.fct_lump_n` are called from both `fct_lump` and their distinct versions, and I extended that pattern for `.fct_lump_min` and `.fct_lump_count` which aren't in `fct_lump` - would you prefer to keep `fct_lump(prop)` as well as `fct_lump_prop` or move the internals into their own functions with a clear separation?

R/lump.R
jonocarroll: It could... I'll change that.

hadley: I think that will solve the `.fct_lump_n` issue above too.

R/lump.R
jonocarroll: Done.

jonocarroll:
Now additionally includes dedicated functions over which `fct_lump` is just a wrapper:

`fct_lump_prop`
`fct_lump_min`
`fct_lump_n`
`fct_lump_count`
`fct_lump_lowfreq`

jonocarroll:
I've reworded some of the 'preserve'/'lump' documentation - it seems `prob` specifies the level at or below which lumping of levels will occur but the help implied it was the value at or above it would be preserved.

hadley:
Overall, looks great — unfortunately I can't review locally because you used a slash in your branch name (which usethis doesn't support). Could you please rename `test-fct_lump.R` to `test-lump.R` and then check the test coverage?

jonocarroll:
I thought `usethis` had support for slashes now? https://github.com/r-lib/usethis/pull/671

jonocarroll:
Moved to https://github.com/tidyverse/forcats/pull/247

PR #245, opened 2020-01-31
add simple example to fct_reorder

### LINKED ISSUE #232: Refresh `fct_reorder()` docs
* [ ] Proofread title, description, and parameters 
* [ ] Proofread comments in examples 
* [ ] Ensure comment headings used to divide into scannable sections 
* [ ] Flag examples that seem overly complex for further review 
* [ ] Switch out mtcars/iris for something more interesting, or a smaller dataset constructed specifically to illustrate one point. 
* [ ] Review size of example outputs to ensure that you can easily see important rows and cols. 
* [ ] Datasets should always be tibbles to avoid use of `as_tibble()` distracting from the main point

johngoldin:
Checked help for fct_reorder. Added a simple data frame at the top of the examples using tribble and then simple examples based on that data. Example is direct use of fct_reorder rather than
using inside another call. I find fct_reoder2 a bit hard to understand and my simple example doesn't help. The ggplot example works better and one is not likely to use fct_reorder alone so perhaps I should not have included the fct_reorder2 example.

Fixes #232

### DIFF
--- a/R/reorder.R
+++ b/R/reorder.R
@@ -20,6 +20,18 @@
 #' @importFrom stats median
 #' @export
 #' @examples
+#' df <- tibble::tribble(
+#' ~color,     ~a, ~b,
+#' "blue",      1,  2,
+#' "green",     6,  2,
+#' "purple",    3,  3,
+#' "red",       2,  3,
+#' "yellow",    5,  1
+#' )
+#' df$color <- factor(df$color)
+#' fct_reorder(df$color, df$a, min)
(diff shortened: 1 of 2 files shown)


R/reorder.R
```diff
@@ -20,6 +20,18 @@
 #' @importFrom stats median
 #' @export
 #' @examples
+#' df <- tibble::tribble(
+#' ~color,     ~a, ~b,
```
hadley: Can you please indent the data rows by two spaces?

PR #268, opened 2020-07-19
Fix partial match in `fct_lump_n()`

### LINKED ISSUE #276: fct_lump_n uses partial argument matching
`fct_lump_n` uses partial argument matching when it calls `rank`, which generates a lot of warnings if you have `options(warnPartialMatchArgs = TRUE)` and repeatedly call the function (e.g. in a grouped operation).

``` r
library(forcats)
options(warnPartialMatchArgs = TRUE)
x <- c("a", "a", "a", "b", "b", "c", "d")
fct_lump_n(x, n = 2)
#> Warning in rank(-calcs$count, ties = ties.method): partial argument match of
#> 'ties' to 'ties.method'
#> [1] a     a     a     b     b     Other Other
#> Levels: a b Other
```

I think the culprit is here, with the named argument given as `ties` instead of `ties.method`.
https://github.com/tidyverse/forcats/blob/ab81d1bf8ef8d1b776dd2a2030e9c664e54df239/R/lump.R#L149-L154

I haven't tried making the change but happy to PR if this would be welcome. I realise `rank` is unlikely to change signature but at least for me squelching the warnings would be great

malcolmbarrett:
Closes #276. This PR fixes a partial argument match in `rank()` in `fct_lump_n()`

``` r
library(forcats)

old <- options(warnPartialMatchArgs = TRUE)
on.exit(options(old), add = TRUE)

x <- factor(letters[rpois(100, 5)])
x <- fct_lump_n(x, n = 6)
#> Warning in rank(-calcs$count, ties = ties.method): partial argument match of
#> 'ties' to 'ties.method'
```

<sup>Created on 2020-07-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

### DIFF
--- a/R/lump.R
+++ b/R/lump.R
@@ -147,10 +147,10 @@ fct_lump_n <- function(f, n, w = NULL, other_level = "Other",
   }
 
   if (n < 0) {
-    rank <- rank(calcs$count, ties = ties.method)
+    rank <- rank(calcs$count, ties.method = ties.method)
     n <- -n
   } else {
-    rank <- rank(-calcs$count, ties = ties.method)
+    rank <- rank(-calcs$count, ties.method = ties.method)
   }
 
   new_levels <- ifelse(rank <= n, levels(f), other_level)


hadley:
Thanks!

PR #272, opened 2020-08-26
Include example that fct_recode can use quotes for new level

lwjohnst86:
(Since it isn't documented). It isn't clear from the documentation that this is possible, so it can be confusing learning how to recode when your new level has spaces or other characters in it (e.g. `-`).


hadley:
This is a general R programming fact, which I'm reluctant to repeat in to many places, so I think it's best to add as a single new example.

PR #302, opened 2022-03-17
Update data.R

PursuitOfDataScience:
`year` does indeed go from 2000 to 2014, but the actual year is every other year, meaning 2000, 2002, 2004, etc. I guess the survey was carried out once in every two years.


hadley:
Thanks!

PR #318, opened 2022-10-02
update link to newer version of R for Data Science

sellisd:
The link to the Factors chapter in the R for Data Science book was pointing towards the first edition. I updated it so it sends to the chapter of the 2nd edition


hadley:
Thanks!

PR #321, opened 2022-11-04
see also: `fct_explicit_na()` <-> `fct_recode()`

dpprdan:
I spent way too much time to day to find the reverse operation of `fct_explicit_na()`, so I thought it might be helpful to reference `fct_recode()` there and vice versa. I think this might be helpful, but feel free to edit or close if you don't agree, of course.


hadley:
I don't love the `"NULL"` interface in `fct_recode()`, so I think we'll tackle this a different way, probably by adding `fct_implicit_na()`

PR #328, opened 2023-01-03
`fct_unique()` improvements

### LINKED ISSUE #284: `fct_unique()` can return values not present in factor
`fct_unique` claims to return unique values of a factor, like `unique` but ordered by the factor levels rather than by order of first appearance. However, for factors in which not all the levels occur at least once, `fct_unique` still returns those levels:

``` r
library(assertthat)
library(forcats)
my_factor <- factor(letters)[1]

unique(my_factor)
#> [1] a
#> Levels: a b c d e f g h i j k l m n o p q r s t u v w x y z
fct_unique(my_factor)
#>  [1] a b c d e f g h i j k l m n o p q r s t u v w x y z
#> Levels: a b c d e f g h i j k l m n o p q r s t u v w x y z
assert_that(all(unique(my_factor) %in% my_factor))
#> [1] TRUE
assert_that(all(fct_unique(my_factor) %in% my_factor))
#> Error: Elements 2, 3, 4, 5, 6, ... of fct_unique(my_factor) %in% my_factor are not true
```

<sup>Created on 2021-08-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

If `fct_unique` is really supposed to be an equivalent of `unique`, it should only return values that actually occur in the factor. Alternatively, it should be documented to return all levels regardless of occurrence and should not be directly compared to `unique` in the documentation.

### LINKED ISSUE #293: `fct_unique()` drops `NA` data if it isn't an explicit level
This came up as an issue in tidyr (https://github.com/tidyverse/tidyr/issues/1275, https://github.com/tidyverse/tidyr/pull/1276/).

``` r
library(forcats)

x <- factor(c(NA, "x"))

fct_unique(x)
#> [1] x
#> Levels: x

# Retains NA data at the end without adding a new factor level
x <- factor(c(NA, "x"))
tidyr:::fct_unique(x)
#> [1] x    <NA>
#> Levels: x

# If an NA factor level already existed, doesn't change the level position
x <- factor(levels = c(NA, "x"), exclude = NULL)
tidyr:::fct_unique(x)
#> [1] <NA> x   
#> Levels: <NA> x
```

We could use the tidyr implementation here:

```r
fct_unique <- function(x) {
   if (!is.factor(x)) {
     abort("`x` must be a factor.")
   }

   levels <- levels(x)
   out <- levels

   if (!anyNA(levels) && anyNA(x)) {
     out <- c(out, NA_character_)
   }

   factor(out, levels = levels, exclude = NULL, ordered = is.ordered(x))
 }
```

hadley:
* Make purpose more clear in docs. Fixes #284.
* Include implicit NA if present. Fixes #293.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # forcats (development version)
 
+* `fct_unique()` now captures includes implicit missing values if present 
+  (#293).
+
 # forcats 0.5.2
 
 * New `fct()` which works like `factor()` but errors if values of `x`
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,6 @@
 # forcats (development version)
 
-* `fct_unique()` now captures includes implicit missing values if present 
-  (#293).
+* `fct_unique()` now captures implicit missing values if present (#293).
 
 # forcats 0.5.2
 
--- a/tests/testthat/test-unique.R
+++ b/tests/testthat/test-unique.R
@@ -8,7 +8,9 @@ test_that("preserve position of explicit NA levels", {
   expect_equal(fct_unique(x), x)
 })
 
-test_that("preserve mix of implicit and explicit NA", {
+test_that("preserve mix of implicit and explicit NA (#328)", {
+  # Possible to create with factor(exclude = NULL) + is.na(), but directly
+  # specifying makes it a little easier to see what's happening.
   f <- structure(
     c(2L, 1L, NA, 1L, 2L, NA),
     levels = c("x", NA),
--- a/tests/testthat/test-unique.R
+++ b/tests/testthat/test-unique.R
@@ -10,13 +10,13 @@ test_that("preserve position of explicit NA levels", {
 
 test_that("preserve mix of implicit and explicit NA", {
   f <- structure(
-    c(2, 1, NA, 1, 2, NA),
+    c(2L, 1L, NA, 1L, 2L, NA),
     levels = c("x", NA),
     class = "factor"
   )
 
   expect_equal(
     fct_unique(f),
-    structure(c(1, 2, NA), levels = c("x", NA), class ="factor")
+    structure(c(1L, 2L, NA), levels = c("x", NA), class = "factor")
   )
 })
--- a/tests/testthat/test-unique.R
+++ b/tests/testthat/test-unique.R
@@ -1,9 +1,22 @@
-test_that("includes missing values", {
+test_that("includes implicit missing values", {
   x <- factor(c(NA, "x"))
   expect_equal(fct_unique(x), fct(c("x", NA)))
 })
 
-test_that("preserves position of implicit NAs", {
+test_that("preserve position of explicit NA levels", {
   x <- fct(c(NA, "x"), levels = c(NA, "x"))
   expect_equal(fct_unique(x), x)
+})
+
+test_that("preserve mix of implicit and explicit NA", {
+  f <- structure(
+    c(2, 1, NA, 1, 2, NA),
+    levels = c("x", NA),
+    class = "factor"
+  )
+
+  expect_equal(
+    fct_unique(f),
+    structure(c(1, 2, NA), levels = c("x", NA), class ="factor")
+  )
 })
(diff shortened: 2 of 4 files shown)


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # forcats (development version)
 
+* `fct_unique()` now captures includes implicit missing values if present 
+  (#293).
```
DavisVaughan, line 4:
```suggestion
* `fct_unique()` now captures implicit missing values if present (#293).
```

tests/testthat/test-unique.R
```diff
@@ -0,0 +1,22 @@
...
+})
+
+test_that("preserve position of explicit NA levels", {
+  x <- fct(c(NA, "x"), levels = c(NA, "x"))
+  expect_equal(fct_unique(x), x)
+})
+
+test_that("preserve mix of implicit and explicit NA", {
+  f <- structure(
+    c(2, 1, NA, 1, 2, NA),
```
DavisVaughan, line 13: should these be integers to be a proper factor?

tests/testthat/test-unique.R
```diff
@@ -0,0 +1,22 @@
...
+  x <- factor(c(NA, "x"))
+  expect_equal(fct_unique(x), fct(c("x", NA)))
+})
+
+test_that("preserve position of explicit NA levels", {
+  x <- fct(c(NA, "x"), levels = c(NA, "x"))
+  expect_equal(fct_unique(x), x)
+})
+
+test_that("preserve mix of implicit and explicit NA", {
```
DavisVaughan, line 11:
```suggestion
test_that("preserve mix of implicit and explicit NA (#328)", {
```

For future readers, we discovered that while most factors replace `NA` with an integer code when `NA` is used as an explicit factor level, it is actually documented in `?factor` that `is.na<-.factor` can be used to simultaneously have `NA` as a code and `NA` as a completely separate explicit factor level that maps to an integer code.

``` r
x <- factor(c("a", NA, NA, "b"), levels = c("a", "b", NA), exclude = NULL)
unclass(x)
#> [1] 1 3 3 2
#> attr(,"levels")
#> [1] "a" "b" NA

is.na(x)[3] <- TRUE

x
#> [1] a    <NA> <NA> b   
#> Levels: a b <NA>
unclass(x) # technically 3 and NA are different
#> [1]  1  3 NA  2
#> attr(,"levels")
#> [1] "a" "b" NA
```

So this test ensures we support that

tests/testthat/test-unique.R
```diff
@@ -0,0 +1,22 @@
...
+  f <- structure(
+    c(2, 1, NA, 1, 2, NA),
+    levels = c("x", NA),
+    class = "factor"
+  )
+
+  expect_equal(
+    fct_unique(f),
+    structure(c(1, 2, NA), levels = c("x", NA), class ="factor")
```
DavisVaughan, line 20:
```suggestion
    structure(c(1L, 2L, NA), levels = c("x", NA), class = "factor")
```

tests/testthat/test-unique.R
DavisVaughan, line 13:
```suggestion
    c(2L, 1L, NA, 1L, 2L, NA),
```

DavisVaughan (approved):

tests/testthat/test-unique.R
DavisVaughan: Fixed in a suggestion below

tests/testthat/test-unique.R
hadley: Pulling this comment out to the top-level so it doesn't get lost.

hadley:
From @DavisVaughan:

For future readers, we discovered that while most factors replace `NA` with an integer code when `NA` is used as an explicit factor level, it is actually documented in `?factor` that `is.na<-.factor` can be used to simultaneously have `NA` as a code and `NA` as a completely separate explicit factor level that maps to an integer code.

``` r
x <- factor(c("a", NA, NA, "b"), levels = c("a", "b", NA), exclude = NULL)
unclass(x)
#> [1] 1 3 3 2
#> attr(,"levels")
#> [1] "a" "b" NA

is.na(x)[3] <- TRUE

x
#> [1] a    <NA> <NA> b   
#> Levels: a b <NA>
unclass(x) # technically 3 and NA are different
#> [1]  1  3 NA  2
#> attr(,"levels")
#> [1] "a" "b" NA
```

So this test ensures we support that

PR #332, opened 2023-01-03
Add `after` argument to `fct_expand()`

### LINKED ISSUE #138: `fct_expand()` needs `after` argument
For `fct_expand`, it would be useful to be able to insert the new level somewhere other than as the last level in the levels.  This is especially true for ordered factors.

Something like the following is what I'm thinking:

``` r
fct_expand <- function(f, ..., after=NULL) {
  f <- forcats:::check_factor(f)
  
  new_levels <- rlang::chr(...)
  if (is.null(after)) {
    all_levels <- union(levels(f), new_levels)
  } else {
    all_levels <-
      append(setdiff(levels(f), new_levels),
             new_levels,
             after=after)
  }
  lvls_expand(f, new_levels=all_levels)
}
```

hadley:
Fixes #138

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # forcats (development version)
 
+* `fct_expand()` gains an `after` argument so that you can choose where
+  the new levels are placed (#138).
+
 # forcats 0.5.2
 
 * New `fct()` which works like `factor()` but errors if values of `x`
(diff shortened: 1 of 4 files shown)


romainfrancois (approved):
LGTM

PR #337, opened 2023-01-05
NA in levels vs NA in values

### LINKED ISSUE #291: `fct_collapse()` should allow `other_level = NA`
IMO the `fct_collapse()` function should support using `NA` as the `other_level` argument.

As of 0.5.0 one cannot use neither `NA` to `NA_character_`
```r
library(purrr)
fac <- c(1, 2, 3, 4) %>% factor()
forcats::fct_collapse(fac, A = c('1', '2'), other_level = NA_character_)
forcats::fct_collapse(fac, A = c('1', '2'), other_level = NA)
```
and the workaround is quite awkward and too long
```r
(
    fac
    %>% forcats::fct_collapse(A = c('1', '2'), other_level = 'NA')
    %>% na_if('NA')
    %>% fct_drop()
)
```

I'd be happy to look into it and make a PR.
This discussion might be relevant https://github.com/tidyverse/forcats/pull/101

### LINKED ISSUE #326: `fct_infreq()` doesn't reorder NA levels
``` r
library(dplyr, warn.conflicts = FALSE)
library(forcats)

starwars |> 
  mutate(hair_color = fct_infreq(hair_color)) |> 
  count(hair_color) |> 
  filter(n > 1)
#> # A tibble: 6 × 2
#>   hair_color     n
#>   <fct>      <int>
#> 1 none          37
#> 2 brown         18
#> 3 black         13
#> 4 white          4
#> 5 blond          3
#> 6 <NA>           5
```

<sup>Created on 2023-01-03 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley:
This is an attempt to take a comprehensive look at handling `NA`s in levels vs `NAs` in values:

* A new terminology that leads to `fct_na_levels_to_values()` and `fct_na_values_to_levels()` (and deprecating `fct_explicit_na()`) 
* A brief introduction to the issue in the forcats vignette (Fixes #326)
* Tests and some small changes to ensure that you can use `other_level = NA` everywhere. (Fixes #291)


R/na.R
```diff
@@ -0,0 +1,57 @@
...
+#' @description
+#' There are two ways to represent missing values in factors: in the values
+#' and in the levels. `NA`s in the values are most useful for data analysis
+#' (since [is.na()] returns what you expect), but because the `NA` is not
+#' explicitly recorded in the levels, there's no way to control it's position
+#' (it's almost always displayed last). Putting the `NA`s in the levels allows
+#' you to control its display, at the cost of losing accurate `is.na()`
+#' reporting.
+#'
+#' (It is possible to have a factor with missing  values in both the values
```
mine-cetinkaya-rundel, line 12:
```suggestion
#' (It is possible to have a factor with missing values in both the values
```

vignettes/forcats.Rmd
```diff
@@ -44,17 +43,59 @@ To do so, we can use the function `fct_infreq()`:
...
+#|   (~35), followed by brown (~18), then black (~12). Surprisingly,
+#|   NAs are at the top of the graph, even though there are ~5 NAs and 
+#|   other colors have smaller values.
+ggplot(starwars, aes(y = fct_infreq(hair_color))) + 
+  geom_bar()
 ```
 
 Note that `fct_infreq()` it automatically puts NA at the top, even though that doesn't have the smallest number of entries.
 
+Its's a little surprising that the `NA` bar isn't ordered by frequency.
```
mine-cetinkaya-rundel, line 58:
```suggestion
It's a little surprising that the `NA` bar isn't ordered by frequency.
```

vignettes/forcats.Rmd
```diff
@@ -44,17 +43,59 @@ To do so, we can use the function `fct_infreq()`:
...
+To understand why, we need to make a brief digression to discuss `NA`s in values vs `NA`s in levels
```
mine-cetinkaya-rundel, line 59:
```suggestion
To understand why we need to make a brief digression to discuss `NA`s in values vs. `NA`s in levels
```

vignettes/forcats.Rmd
```diff
@@ -44,17 +43,59 @@ To do so, we can use the function `fct_infreq()`:
...
+
+-   You can include it in the values of the factor; it does not appear in the levels and `is.na()` reports it as missing. This is the usual way to encode a missing value in a factor:
+
+```{r}
+f <- factor(c("x", "y", NA))
+levels(f)
+is.na(f)
+```
+
+-   You can include in the levels of the factor and `is.na()` does not report it as missing. This requires a little more work to create:
```
mine-cetinkaya-rundel, line 73:
```suggestion
-   You can include it in the levels of the factor and `is.na()` does not report it as missing. This requires a little more work to create:
```

vignettes/forcats.Rmd
```diff
@@ -44,17 +43,59 @@ To do so, we can use the function `fct_infreq()`:
...
 Note that `fct_infreq()` it automatically puts NA at the top, even though that doesn't have the smallest number of entries.
 
+Its's a little surprising that the `NA` bar isn't ordered by frequency.
+To understand why, we need to make a brief digression to discuss `NA`s in values vs `NA`s in levels
+
+## `NAs`s in levels and values
+
+There are two ways to represent a missing value in a factor:
+
+-   You can include it in the values of the factor; it does not appear in the levels and `is.na()` reports it as missing. This is the usual way to encode a missing value in a factor:
```
mine-cetinkaya-rundel, line 65: It's not just the "usual" but also the "default", so maybe "This is how missing values are encoded in a factor by default"?

vignettes/forcats.Rmd
```diff
@@ -44,17 +43,59 @@ To do so, we can use the function `fct_infreq()`:
...
+```{r}
+f <- factor(c("x", "y", NA))
+levels(f)
+is.na(f)
+```
+
+-   You can include in the levels of the factor and `is.na()` does not report it as missing. This requires a little more work to create:
+
+```{r}
+f <- factor(c("x", "y", NA), exclude = NULL)
```
mine-cetinkaya-rundel, line 76: Is there a way to say "don't exclude NAs" without specifying `exclude = NULL`? If not, I think it would be good to add a sentence about why we brought `NULL` into play all of a sudden, like something about "this works because in R `NULL`s are ..."

vignettes/forcats.Rmd
```diff
@@ -44,17 +43,59 @@ To do so, we can use the function `fct_infreq()`:
...
+levels(f)
+is.na(f)
+```
+
+`NA`s in the values tend to be best for data analysis, since `is.na()` works as you'd expect.
+`NA`s in the levels are often better for display since you can control where they appear.
```
mine-cetinkaya-rundel, line 82:
```suggestion
`NA`s in the levels can be useful for display if you need to control where they appear in a table or a plot.
```

vignettes/forcats.Rmd
```diff
@@ -44,17 +43,59 @@ To do so, we can use the function `fct_infreq()`:
...
+
+to fix the issue above, we can use `fct_na_value_to_level()` to convert the `NA` in the value to an NA in the levels.
```
mine-cetinkaya-rundel, line 84:
```suggestion
To fix the issue above, we can use `fct_na_value_to_level()` to convert the `NA` in the value to an NA in the levels.
```

vignettes/forcats.Rmd
mine-cetinkaya-rundel, line 82: I would move these two sentences to the end of this section since they're like a summary.

Also, I wonder about saying something like 

> You should be careful about using `fct_na_value_to_level()` in a data transformation pipeline where you overwrite your data since that will affect all downstream analysis as well. This should only be done if you will be using that updated data frame or creating displays or summaries, chances are you don't want NAs as an explicit level in your data analysis. If you do, you might consider relabeling them as something like "Unknown" or "Not available".

Maybe this is an overreach, but I feel like this function should be used as a quick fix, but if you really care about those `NA`s it's more practical to not make them `NA`s but something else, so they don't accidentally get swooped up by another function that treats NAs in a special way, say in a modelling context or something...

mine-cetinkaya-rundel (changes requested):
I like making it more explicit (ha!) what "explicit" means so moving from value->level / level->value language from "explicit NA" sounds good to me.

I left a few typo fixes as well as some suggestions for how to present this idea in the vignette.

vignettes/forcats.Rmd
DavisVaughan: > If you do, you might consider relabeling them as something like "Unknown" or "Not available".

Which you'd do with `fct_na_value_to_level(x, name = "Unknown")` if we add the idea of `fct_explicit_na(na_level =)` back in

vignettes/forcats.Rmd
```diff
@@ -44,17 +43,59 @@ To do so, we can use the function `fct_infreq()`:
...
+Then they appear where you'd expect:
+
+```{r}
+#| fig.alt: >
+#|   The bar chart of hair color, now ordered so that NAs, now labelled
+#|   "(missing)" are ordered where you'd expect: in between white (4) and
```
DavisVaughan, line 90: Narrator: "They were not, in fact, labelled `"(missing)"`"

DavisVaughan (approved):

R/na.R
```diff
@@ -0,0 +1,57 @@
...
+#' # If needed, you can convert back to NAs in the values:
+#' f3 <- fct_na_level_to_value(f2)
+#' levels(f3)
+#' as.integer(f3)
+#' is.na(f3)
+fct_na_value_to_level <- function(f) {
+  f <- check_factor(f)
+
+  f <- fct_expand(f, NA)
+  f[is.na(f)] <- NA
```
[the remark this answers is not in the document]
hadley: This is like `f[is.na(f)] <- "x"` where the RHS is matched into the levels (which now contain `NA`) because of the previous line.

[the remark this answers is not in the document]
hadley: Oooooh, I didn't notice that. Yes, you're right that it's unnecessary.

R/collapse.R
```diff
@@ -31,19 +31,21 @@ fct_collapse <- function(.f, ..., other_level = NULL, group_other = "DEPRECATED"
...
-    new[[other_level]] <- levels[!levels %in% levs]
-    levs <- c(levs, new[[other_level]])
+    not_specified <- setdiff(levels(f), old)
+
+    old <- c(old, not_specified)
+    new <- c(new, rep(other_level, length(not_specified)))
   }
 
-  names(levs) <- names(new)[rep(seq_along(new), vapply(new, length, integer(1)))]
-  out <- fct_recode(.f, !!!levs)
```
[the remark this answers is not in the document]
hadley: I added https://github.com/tidyverse/forcats/issues/339 to think about this more. I can't figure out if it deserves a test or not. On the one hand, it was arguably a bug, so it should get a test. On the other hand, thinking about `fct_collapse()` from first principles, it seems strange to test specific values of `other_level`.

R/other.R
```diff
@@ -18,17 +18,22 @@ fct_other <- function(f, keep, drop, other_level = "Other") {
...
-    levels[levels %in% drop] <- other_level
+    lvls_other(f, !levels(f) %in% drop, other_level)
   }
+}
 
-  if (!other_level %in% levels) {
-    return(f)
+# Replace specified levels (if any), with other.
+# @param keep A logical vector the same length as `levels(f)`
+lvls_other <- function(f, keep, other_level = "Other") {
```
hadley, line 30: I also added this new helper function to ensure that the implementation of "othering" is identical for all code paths.

R/na.R
```diff
@@ -0,0 +1,72 @@
...
+  new_levels[is.na(new_levels)] <- level
+
+  lvls_revalue(f, new_levels)
+}
+
+#' @export
+#' @rdname fct_na_value_to_level
+#' @param extra_levels Optionally, a character vector giving additional levels
+#'   that should also be converted to `NA` values.
+fct_na_level_to_value <- function(f, extra_levels = NULL) {
```
hadley, line 57: I made this `extra_levels` because if you don't include `NA`, it's easy to create factors that have a mix of NAs in values and levels.

R/explicit_na.R
```diff
@@ -1,21 +1,40 @@
...
+#' # previously
 #' f2 <- fct_explicit_na(f1)
+#' # now
+#' f2 <- fct_na_value_to_level(f1)
+#'
 #' fct_count(f2)
-#' table(is.na(f2))
+#' table(f2)
+#' sum(is.na(f2))
 fct_explicit_na <- function(f, na_level = "(Missing)") {
```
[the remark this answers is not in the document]
hadley: Take a look below.

R/na.R
```diff
@@ -33,22 +35,37 @@
 #' levels(f3)
 #' as.integer(f3)
 #' is.na(f3)
-fct_na_value_to_level <- function(f) {
+fct_na_value_to_level <- function(f, level = NA) {
+  if (!identical(level, NA) && !is_string(level)) {
+    cli::cli_abort(
+      "{.arg level} must be a string or {.code NA}, not {.obj_type_friendly level}."
+    )
+  }
```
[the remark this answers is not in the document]
DavisVaughan: Oh I guess you aren't using the checkers yet

R/na.R
```diff
@@ -33,22 +35,37 @@
 #' levels(f3)
 #' as.integer(f3)
 #' is.na(f3)
-fct_na_value_to_level <- function(f) {
+fct_na_value_to_level <- function(f, level = NA) {
+  if (!identical(level, NA) && !is_string(level)) {
+    cli::cli_abort(
+      "{.arg level} must be a string or {.code NA}, not {.obj_type_friendly level}."
```
DavisVaughan, line 41: I think you need `{.obj_type_friendly {level}}`

It isn't letting me do a "suggestion" for some reason

R/na.R
```diff
@@ -33,22 +35,37 @@
...
+  }
   f <- check_factor(f)
 
-  # re-matches values to levels
-  fct_expand(f, NA)
+  f <- fct_expand(f, NA)
+  new_levels <- levels(f)
+  new_levels[is.na(new_levels)] <- level
+
+  lvls_revalue(f, new_levels)
```
DavisVaughan, line 50: So if you already have an `NA` level, this will rename it to `level`? Seems reasonable

R/na.R
DavisVaughan: Makes sense. I was actually wondering if we even needed `extra_levels` at all, but there doesn't seem to be any other way to do this in forcats, so it seems useful to have

``` r
x <- factor(c("x", "missing"))

forcats::fct_na_level_to_value(x, extra_levels = "missing")
#> [1] x    <NA>
#> Levels: x
```

It is a nice way of declaring, "no, `missing` should really be a missing value"

R/na.R
```diff
@@ -33,22 +35,37 @@
...
 #' @export
 #' @rdname fct_na_value_to_level
-fct_na_level_to_value <- function(f) {
+#' @param extra_levels Optionally, a character vector giving additional levels
+#'   that should also be converted to `NA` values.
+fct_na_level_to_value <- function(f, extra_levels = NULL) {
   f <- check_factor(f)
+  if (!is.null(extra_levels) && !is.character(extra_levels)) {
+    cli::cli_abort(
+      "{.arg extra_levels} must be a string or {.code NULL}, not {.obj_type_friendly extra_levels}."
```
DavisVaughan, line 61: Same comment about `{.obj_type_friendly {extra_levels}}`

DavisVaughan (approved):

R/na.R
DavisVaughan: I'm actually surprised this has never come up before. It seems pretty useful

R/na.R
hadley: Doh

PR #341, opened 2023-01-09
Improved handling of NAs in `fct_reorder()`

### LINKED ISSUE #266: `fct_reorder()` needs `na_last` argument
It is not currently possible to easily make a horizontal bar graph where bars are ordered by size and with NA bars on the bottom of the graph. 

The graph looks good vertically with NAs on the far right, but as soon as you flip the plot, the NAs are at the top, which in general is not good, given that you almost never want to emphasise the NA values.

A solution to this might be to update forcats::fct_reorder to have an argument to support this (e.g. na_flip = TRUE).

```r
library(dplyr)
library(ggplot2)

plot_data <- iris %>% 
  group_by(Species) %>% 
  summarise(Petal.Length = mean(Petal.Length)) %>% 
  add_row(Species = "New species") %>% 
  mutate(Species = forcats::fct_reorder(Species, Petal.Length))

ggplot(plot_data) +
  geom_col(aes(Species, Petal.Length))
```
image

```
ggplot(plot_data) +
  geom_col(aes(Species, Petal.Length)) + 
  coord_flip()
```

image

### LINKED ISSUE #315: `fct_reorder()` yields unexpected result in presence of missing values
Hi,

When using `fct_reorder()` in presence of missing values, you often do not get the expected result.

For instance, in the following code, the "blue" level gets an `NA` summary and is therefore sent to the last level of the result. In larger datasets, where missing values happen everywhere, this results in `fct_reorder()` doing frustratingly nothing.

``` r
library(tidyverse)
df = tribble(
  ~color,     ~a,
  "purple",      1,
  "purple",      2,
  "blue",     3, #NA
  "blue",     4,
  "green",    5,
  "green",    6
)
df$color = factor(df$color)
df$color %>% levels
#> [1] "blue"   "green"  "purple"
fct_reorder(df$color, df$a) %>% levels
#> [1] "purple" "blue"   "green"

df$a[3]=NA
fct_reorder(df$color, df$a) %>% levels
#> [1] "purple" "green"  "blue"
fct_reorder(df$color, df$a, na.rm=TRUE) %>% levels
#> [1] "purple" "blue"   "green"
```

<sup>Created on 2022-08-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This is especially unexpected as the default function, `median()`, has `na.rm=FALSE` by default. Using other common summary functions like `min()` and `max()` has the same problem.

There is a mention of this in the documentation (`...  Other arguments passed on to .fun. A common argument is na.rm = TRUE.`), but I don't think this is explicit enough.

Could there be some kind of warning to suggest we add `na.rm=TRUE`? For instance `if(any(is.na(summary))) warn("missing")`.

Otherwise, maybe this should be mentioned up in the description, for instance something like "Any missing value returned by the summary function for a level will cause this level to be sent to the end." (ok that's not well written but you get the point)

You might even want the user to explicitly opt-in for `na.rm=FALSE`, and by default inject `na.rm=TRUE` to the summary function if `na.rm` is in `formals(.fun).`. This is a bit invasive, I'll give you that, but I cannot see any real use case where `na.rm=FALSE` could be wanted.

hadley:
* Drop missing values in `.x` by default. Fixes #315.
* New `.na_last` argument to control position of NAs. Fixes #266.

@DavisVaughan what do you think of this API? It's a bit unusual to default to `na.rm = TRUE`, but you can't see the result of the computation and including `NA`s is never useful, so adopting ggplot2's approach makes sense to me. Once you've reviewed this API, I'll also apply it to `fct_reorder2()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # forcats (development version)
 
+* `fct_reorder()` now removes `NA` values in `.x` with a warning (like
+   `ggplot2::geom_point()` and friends). You can suppress the warning by
+   setting `.na_rm = TRUE` (#315).
+   
+* `fct_reorder()` gains a new `.na_last` argument that allows you to control
+  where levels that are empty or only contain missing values are placed (#266).
+
 * `fct_explicit_na()` is deprecated in favour of `fct_na_value_to_level()`.
 
 * New `fct_na_value_to_level()` and `fct_na_level_to_value()` to convert 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,8 +4,9 @@
    `ggplot2::geom_point()` and friends). You can suppress the warning by
    setting `.na_rm = TRUE` (#315).
    
-* `fct_reorder()` gains a new `.na_last` argument that allows you to control
-  where levels that are empty or only contain missing values are placed (#266).
+* `fct_reorder()` gains a new `.default` argument controls the placement of 
+  empty levels (including levels that might become empty after removing
+  missing values in `.x`) (#266).
 
 * `fct_explicit_na()` is deprecated in favour of `fct_na_value_to_level()`.
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
    `ggplot2::geom_point()` and friends). You can suppress the warning by
    setting `.na_rm = TRUE` (#315).
    
-* `fct_reorder()` gains a new `.default` argument controls the placement of 
+* `fct_reorder()` gains a new `.default` argument that controls the placement of 
   empty levels (including levels that might become empty after removing
   missing values in `.x`) (#266).
 
--- a/R/reorder.R
+++ b/R/reorder.R
@@ -12,7 +12,7 @@
 #' @param .fun n summary function. It should take one vector for
 #'   `fct_reorder`, and two vectors for `fct_reorder2`, and return a single
 #'   value.
-#' @param .na_rm Should `fct_reorder()` silently remove missing values?
+#' @param .na_rm Should `fct_reorder()` remove missing values?
 #'   If `NULL`, the default, will remove missing values with a warning.
 #'   Set to `FALSE` to preserve `NA`s (if you `.fun` already handles them) and
 #'   `TRUE` to remove silently.
--- a/R/reorder.R
+++ b/R/reorder.R
@@ -13,10 +13,11 @@
 #'   `fct_reorder`, and two vectors for `fct_reorder2`, and return a single
 #'   value.
 #' @param .na_rm Should `fct_reorder()` silently remove missing values?
-#'   If `FALSE`, the default, will remove missing values with a warning.
-#' @param .na_last Should `NA` values be ordered last (`TRUE`) or first
-#'   (`FALSE`). Particularly important for empty levels, since their
-#'   summary result will always be `NA`.
+#'   If `NULL`, the default, will remove missing values with a warning.
+#'   Set to `FALSE` to preserve `NA`s and `TRUE` to remove silently.
+#' @param .default What default value should we use for `.fun` for
+#'   empty levels? Use this to control where empty levels appear in the
+#'   output.
 #' @param ... Other arguments passed on to `.fun`.
 #' @param .desc Order in descending order? Note the default is different
 #'   between `fct_reorder` and `fct_reorder2`, in order to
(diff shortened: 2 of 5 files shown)


R/reorder.R
```diff
@@ -45,15 +49,24 @@
 #'   geom_point() +
 #'   geom_line() +
 #'   labs(colour = "Chick")
-fct_reorder <- function(.f, .x, .fun = median, ..., .desc = FALSE) {
+fct_reorder <- function(.f, .x, .fun = median, ..., .na_rm = FALSE, .na_last = TRUE, .desc = FALSE) {
   f <- check_factor(.f)
   stopifnot(length(f) == length(.x))
   check_dots_used()
 
+  miss <- is.na(.x)
```
DavisVaughan, line 57: FWIW I also did some checking to make sure that it currently works as expected if there are any missing values or missing levels in `.f`. It seems fine, so this PR is definitely just focused on `.x`.

NEWS.md
```diff
@@ -1,5 +1,12 @@
 # forcats (development version)
 
+* `fct_reorder()` now removes `NA` values in `.x` with a warning (like
+   `ggplot2::geom_point()` and friends). You can suppress the warning by
+   setting `.na_rm = TRUE` (#315).
+   
+* `fct_reorder()` gains a new `.na_last` argument that allows you to control
+  where levels that are empty or only contain missing values are placed (#266).
```
DavisVaughan, line 8: This description doesn't seem quite right. I don't think the `NA`s appear in the _levels_, they appear in `.x`.

If you keep the current `.na_rm` behavior, is this really ever used?

tests/testthat/test-reorder.R
```diff
@@ -71,6 +71,21 @@ test_that("complains if summary doesn't return single value", {
...
+
+  expect_snapshot(f2 <- fct_reorder(f1, x))
+  expect_equal(levels(f2), c("c", "b", "a"))
+})
+
+test_that("can control the placement of NA levels", {
+  f1 <- fct(c("a", "b", "c"), letters[1:4])
+  x <- c(1, 2, 3)
+
+  f2 <- fct_reorder(f1, x, .na_last = FALSE)
```
DavisVaughan, line 86: Ah I see. The reason this kicks in is because `tapply(default = NA)` kicks in for the `"d"` level, which doesn't have a `.x` value corresponding to it. I wouldn't say that is an "NA level".

In fact, the fact that this argument controls the placement of the `"d"` level _at all_ is kind of confusing if you don't know the internals and the `tapply()` usage. There are no `NA` values in `f1` or `x`, so it seems weird that `.na_last` does anything.

Maybe it would be clearer if `.default = NA` was an exposed argument to `fct_reorder()`? That would:
- Make it clear that `.na_last` affects the `.default` case if you leave it as `NA`
- Let users set `.default = 0` to make `.na_last` only affect where `.x` had missing values and they set `.na_rm = FALSE` explicitly.

NEWS.md
DavisVaughan: I now see where this is used down in the tests. See more comments there.

But I still think the description isn't quite right. It's like it is used for:
- Levels that don't appear in the data (and `.default = NA`)
- When `.x` has missing values (and `.na_rm = FALSE`)

R/reorder.R
```diff
@@ -45,15 +49,24 @@
...
+  if (any(miss) && !isTRUE(.na_rm)) {
+    cli::cli_warn(c(
+      "{.fn fct_reorder} removing {sum(miss)} missing value{?s}.",
+      i = "Use `.na_rm = TRUE` to silence this message"
+    ))
+  }
```
DavisVaughan, line 63: I'm slightly uncomfortable with the fact that you can't _not remove_ missing values, i.e. if you expect your summary function to handle them somehow. What about:

- `.na_rm = NULL`. Does `TRUE` with a warning if there are `NA`s in `.x`. This is the default. Warning encourages explicitly setting `.na_rm`.
- `.na_rm = FALSE`. Does nothing.
- `.na_rm = TRUE`. Does `TRUE` silently.

DavisVaughan (commented):
My two concrete suggestions from the comments below are:
- `.na_rm = NULL` as the default, which does `TRUE` with a warning if there are `NA`s, but also allows you to set it to `FALSE` for the CRAN behavior
- Expose `.default = NA` passed through to `tapply()`. This makes it clear that `.na_last` applies to empty levels and gives the user a way to opt-out of having `.na_last` affect empty levels by setting `.default = 0` (or something like that) as needed.

tests/testthat/test-reorder.R
hadley: I like the idea of exposing `.default` instead of `.na_last`, that gives you the same control, but it's less confusing.

hadley:
@DavisVaughan using `.default` feels a lot nicer, and I like the idea to default `.na_rm` to `NULL`.

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # forcats (development version)
 
+* `fct_reorder()` now removes `NA` values in `.x` with a warning (like
+   `ggplot2::geom_point()` and friends). You can suppress the warning by
+   setting `.na_rm = TRUE` (#315).
+   
+* `fct_reorder()` gains a new `.default` argument controls the placement of
```
DavisVaughan, line 7:
```suggestion
* `fct_reorder()` gains a new `.default` argument that controls the placement of 
```

R/reorder.R
```diff
@@ -12,8 +12,14 @@
 #' @param .fun n summary function. It should take one vector for
 #'   `fct_reorder`, and two vectors for `fct_reorder2`, and return a single
 #'   value.
-#' @param ... Other arguments passed on to `.fun`. A common argument is
-#'   `na.rm = TRUE`.
+#' @param .na_rm Should `fct_reorder()` silently remove missing values?
```
DavisVaughan, line 15:
```suggestion
#' @param .na_rm Should `fct_reorder()` remove missing values?
```

R/reorder.R
```diff
@@ -45,14 +51,31 @@
 #'   geom_point() +
 #'   geom_line() +
 #'   labs(colour = "Chick")
-fct_reorder <- function(.f, .x, .fun = median, ..., .desc = FALSE) {
+fct_reorder <- function(.f, .x, .fun = median, ..., .na_rm = NULL, .default = Inf, .desc = FALSE) {
```
DavisVaughan, line 54: Want to do any `check_bool(.na_rm, allow_null = TRUE)` style checks?

Seems reasonable to not check `.default`

tests/testthat/test-reorder.R
```diff
@@ -71,6 +71,26 @@ test_that("complains if summary doesn't return single value", {
...
+  expect_no_warning(f3 <- fct_reorder(f1, x, .na_rm = FALSE))
+  expect_equal(levels(f3), c("b", "a", "c"))
+})
+
+test_that("can control the placement of empty levels", {
+  f1 <- fct(c("a", "b", "c"), letters[1:4])
+  x <- c(1, 2, 3)
+
+  f2 <- fct_reorder(f1, x, .default = -Inf)
+  expect_equal(levels(f2), c("d", "a", "b", "c"))
```
DavisVaughan, line 92: Might also want a test where `x` is `x <- c(1, 2, 3, NA)` and you use `.default = -Inf` to show that `.default` kicks in in cases where the `.na_rm` option removed all values corresponding to a particular level (here `"d"`)

DavisVaughan (approved):

R/reorder.R
hadley: Yeah, but I need to pull in the rlang helpers, so I'll do that in another PR.

PR #344, opened 2023-01-10
Add rlang style argument checking to all functions

hadley:


R/collapse.R
```diff
@@ -19,6 +19,7 @@
 #' fct_count(partyid2)
 fct_collapse <- function(.f, ..., other_level = NULL, group_other = "DEPRECATED") {
   f <- check_factor(.f)
+  check_string(other_level, allow_null = TRUE, allow_na = TRUE)
```
DavisVaughan, line 22: Do we allow empty strings for levels?

R/expand.R
```diff
@@ -15,6 +15,7 @@
 fct_expand <- function(f, ..., after = Inf) {
   f <- check_factor(f)
   check_dots_unnamed()
+  check_number_decimal(after)
```
DavisVaughan, line 18: Minimum value of `0` i believe?

Also is it `check_number_whole()`?

R/lump.R
```diff
@@ -87,25 +87,21 @@ fct_lump <- function(f, n, prop, w = NULL, other_level = "Other",
 #' @rdname fct_lump
 fct_lump_min <- function(f, min, w = NULL, other_level = "Other") {
   f <- check_factor(f)
-  level_w <- compute_weights(f, w)
-
-  if (!is.numeric(min) || length(min) != 1 || min < 0) {
-    cli::cli_abort("{.arg min} must be a positive number.")
-  }
+  check_number_decimal(min, min = 0)
```
DavisVaughan, line 90: whole vs decimal?

R/shift.R
```diff
@@ -18,6 +18,7 @@
 #' fct_shift(x, 2)
 #' fct_shift(x, -1)
 fct_shift <- function(f, n = 1L) {
```
DavisVaughan, line 20: `check_factor()` here?

DavisVaughan (approved):

R/collapse.R
hadley: Hmmm, technically I guess we should.

R/expand.R
hadley: `check_number_whole(Inf)` throws an error 😞

Issue #83, opened 2017-02-10
Preserve attributes in fct_drop?


dpavlic:
One of the ways in which fct_drop is very much the same as droplevels is that it removes attributes (such as Hmisc labels) when it drops extra factor levels:

```
library(Hmisc)
library(forcats)

a <- data.frame(a = factor(c('a', 'b', 'c')), b = c(1:3))
a <- a[1:2, ]
label(a$a) <- 'Var a'
a$a_drop <- fct_drop(a$a)
```

They are there in the original:

```
> attributes(a$a)
$levels
[1] "a" "b" "c"

$class
[1] "labelled" "factor"  

$label
[1] "Var a"
```

Gone in the drop:

```
> attributes(a$a_drop)
$levels
[1] "a" "b"

$class
[1] "factor"
```

Would it be possible to preserve them?

hadley:
I can't see any easy way to make this happen without a bunch of work unfortunately. `fct_drop()` has to create a new factor, and there's no obviously way to make this generic.

hadley:
Hmmm, maybe this is the point of the `refactor()` function

hadley:
Fixed in 090160d108ce589eabf86a4cda50e09a85b80552

tjtnew:
I think this is still an issue as `refactor` overwrites the class with the reset result of `factor`:  
https://github.com/tidyverse/forcats/blob/fb11604d2e67f9ef325b1bebded93244dedcfb15/R/lvls.R#L92-L100 I think this can be fixed by adding `class(new_f) <- NULL` in between lines 97 and 98.  Happy to do a pull request if appropriate.  

Example of current behaviour below
``` r
library(forcats, quietly = TRUE)
library(Hmisc, quietly = TRUE)

a <- data.frame(a = factor(c('a', 'b', 'c')), b = c(1:3))
a <- a[1:2, ]
label(a$a) <- 'Var a'
a$a_drop <- fct_drop(a$a)

# original attributes
attributes(a$a)
#> $levels
#> [1] "a" "b" "c"
#> 
#> $class
#> [1] "labelled" "factor"  
#> 
#> $label
#> [1] "Var a"

# new attributes - labelled class dropped
attributes(a$a_drop)
#> $levels
#> [1] "a" "b"
#> 
#> $class
#> [1] "factor"
#> 
#> $label
#> [1] "Var a"
```

khusmann:
[...]

PR #367, opened 2024-08-09
preserve subclasses as well as attrs in `refactor()`

khusmann:
As @TimTaylor first pointed out in #83, `refactor()` overwrites the class with the reset result of factor. This affects a bunch of methods downstream of `refactor()` like lvls_expand(), lvls_reorder(), fct_lump_n(), etc.

Because `refactor()` sometimes needs to set `ordered` in the class list, it's not as simple as ignoring the class of the new factor. Instead, if `ordered` is being toggled off, it drops "ordered" from the original class list; if it's being toggled on, it adds "ordered" to the original class list; if nothing is changed it uses the original class list. As a result, all the subclasses in the original factor propagate to the result, instead of getting lost.

The tests in `lvls-test.R` demonstrate this functionality.

Cheers!

### DIFF
--- a/R/lvls.R
+++ b/R/lvls.R
@@ -95,6 +95,17 @@ refactor <- function(f, new_levels, ordered = NA) {
 
   new_f <- factor(f, levels = new_levels, exclude = NULL, ordered = ordered)
   attributes(new_f) <- utils::modifyList(attributes(f), attributes(new_f))
+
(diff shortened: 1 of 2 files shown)


hadley:
Thank you!

khusmann:
Thank you for the merge!

Can't wait to submit more for dev day, this'll be the first year I've been able to attend. :)

hadley:
Awesome!

PR #395, opened 2026-06-06
fix: fct_reorder2() drops values when .na_rm removes NAs (#348)

### LINKED ISSUE #348: `fct_reorder2()` drops values
``` r
library(forcats)
library(dplyr, warn.conflicts = FALSE)

by_age <- gss_cat |>
  count(age, marital) |>
  group_by(age) |>
  mutate(
    prop = n / sum(n)
  )

nrow(by_age)
#> [1] 357
with(by_age, length(fct_reorder2(marital, age, prop, .na_rm = TRUE)))
#> [1] 351
with(by_age, length(fct_reorder2(marital, age, prop)))
#> Warning: `fct_reorder2()` removing 6 missing values.
#> ℹ Use `.na_rm = TRUE` to silence this message.
#> ℹ Use `.na_rm = FALSE` to preserve NAs.
#> [1] 351
```

<sup>Created on 2023-02-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

LeonidasZhak:
Fixes #348

## Problem

When `fct_reorder2()` is called with `.na_rm = TRUE` (or when `.na_rm = NULL` and NAs trigger the default removal with a warning), the returned factor is shorter than the input. This causes downstream errors like:

```
Error in `mutate()`:
! `m` must be size 357 or 1, not 351.
```

The bug is that `.f` is subsetted along with `.x` and `.y` during NA removal, then the shortened `.f` is passed to `lvls_reorder()`.

## Fix

Save the original `.f` before NA filtering and use it in `lvls_reorder()`. This is the same pattern already applied to `fct_reorder()` in PR #394.

**R/reorder.R**: 2 lines changed (save `f <- .f`, use `f` in `lvls_reorder()`)

## Tests

Added test verifying `fct_reorder2()` preserves input length when NAs are present in `.x` or `.y`.

## Validation

All existing tests pass. Reproduced the original bug scenario from the issue (357 input rows, 351 output before fix, 357 after fix).

### DIFF
--- a/R/reorder.R
+++ b/R/reorder.R
@@ -103,6 +103,7 @@ fct_reorder2 <- function(
   .desc = TRUE
 ) {
   .f <- check_factor(.f)
+  f <- .f
   stopifnot(length(.f) == length(.x), length(.x) == length(.y))
   check_dots_used()
   check_bool(.na_rm, allow_null = TRUE)
@@ -134,7 +135,7 @@ fct_reorder2 <- function(
   )
   check_single_value_per_group(summary, ".fun")
 
-  lvls_reorder(.f, order(summary, decreasing = .desc))
+  lvls_reorder(f, order(summary, decreasing = .desc))
 }
 
 check_single_value_per_group <- function(x, fun_arg, call = caller_env()) {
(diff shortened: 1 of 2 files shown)


LeonidasZhak:
I am withdrawing this PR as part of a broader cleanup of an oversized automated contribution batch. I am sorry for the review noise and the pressure this may have put on maintainers. Going forward I will keep contributions to this project much more limited and higher-signal, ideally only one or two focused PRs at a time. Thank you for maintaining the project.

PR #399, opened 2026-06-06
test: add edge-case tests for fct_rev()

LeonidasZhak:
## Summary

Adds edge-case test coverage for `fct_rev()`, which previously had only 1 test.

## Tests Added

1. **NA values**: Verifies NA positions are preserved after level reversal
2. **Empty factor**: Verifies empty input returns empty factor with no levels
3. **Single level**: Verifies single-level factor is unchanged
4. **Unused levels**: Verifies unused levels are reversed correctly
5. **Ordered class**: Verifies ordered factor class is preserved

## Tests Run

```
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 10 ]
```

All 10 tests in `test-rev.R` pass (1 existing + 5 new). Full suite: 229 PASS, 0 FAIL.

### DIFF
--- a/tests/testthat/test-rev.R
+++ b/tests/testthat/test-rev.R
@@ -4,3 +4,42 @@ test_that("reverses levels", {
 
   expect_equal(levels(f2), c("b", "a"))
 })
+
+test_that("preserves NA values in data", {
+  f1 <- factor(c("a", "b", NA, "c"))
+  f2 <- fct_rev(f1)
+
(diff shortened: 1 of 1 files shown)


LeonidasZhak:
Withdrawing this small automated PR while I consolidate an oversized batch of contributions and reduce maintainer review burden. Sorry for the noise, and thank you for maintaining the project.

